PR #149, opened 2021-06-15
BH-982: specify media types for media input

matthewguywright:


nickcernis (commented):
This is looking great, @matthewguywright! Nice work.

I'm ok to defer restricting uploadable file types for now if the team is happy with that. From Slack it looks like we'd have to build a custom component to restrict upload types. I think it's enough to restrict file types in the media picker so other types cannot be selected, as you've done here.

The acceptance criteria say, “…publishers receive helpful feedback about accepted media types…”.

Right now they can't really tell what types are accepted and have to figure it out from what appears in the media picker. If no files are in the media area they won't know what types are accepted.

**Do you think we should provide a clue below the field title in the publisher app to help with this?** Maybe we could list allowed file types (if there are any) on the same line as the “required” field? For example, by editing this line (you will need to pull and merge main to see it):

https://github.com/wpengine/atlas-content-modeler/blob/e9becf67924f7fd9c49cec30154c6694d5f8e782/includes/publisher/js/src/components/MediaUploader.jsx#L67

So users would see something like:

```
*Required. Accepts file types: png, jpg, jpeg.
```

Another place we could display it is the Media picker title bar:

Screenshot 2021-06-16 at 13 54 14

Screenshot 2021-06-16 at 13 54 06

For example:

```diff
diff --git a/includes/publisher/js/src/components/MediaUploader.jsx b/includes/publisher/js/src/components/MediaUploader.jsx
index db45c3a..c5e5798 100644
--- a/includes/publisher/js/src/components/MediaUploader.jsx
+++ b/includes/publisher/js/src/components/MediaUploader.jsx
@@ -79,8 +79,17 @@ export default function MediaUploader({ modelSlug, field, required }) {
 			library.type = getAllowedTypesLongExtension();
 		}
 
+		const getMediaModalTitle = () => {
+			const title = mediaUrl ? "Change Media" : "Upload Media";
+			if (allowedTypes) {
+				return `${title}  (${allowedTypes.split(",").join(", ")})`;
+			}
+
+			return title;
+		};
+
 		const media = wp.media({
-			title: mediaUrl ? "Change Media" : "Upload Media",
+			title: getMediaModalTitle(),
 			multiple: false,
 			frame: "select",
 			library: library,
```

Adding labels like these also gives us an easier way of writing e2e tests that check that changes made in the developer app affect the publisher app.

ChrisWiegman (changes requested):
Found some more i18n issues that need addressing (sorry, realize this wasn't in place when you started the PR but we do need to keep these up to date going forward)

ChrisWiegman (dismissed):

nickcernis (dismissed):

mindctrl (approved):

mindctrl (dismissed):
Just realized this doesn't have tests.

includes/settings/js/src/components/fields/AdvancedSettings.jsx
```diff
@@ -5,6 +5,90 @@ import React from "react";
...
+				)}
+			</p>
+
+			<p className="mb-4">
+				<a
+					href="https://wpengine.com/support/mime-types-wordpress/"
+					target="_blank"
+					rel="noreferrer"
+				>
+					{__("Wordpress Allowed Types", "atlas-content-modeler")}
```
mindctrl, line 31:
```suggestion
					{__("WordPress Allowed Types", "atlas-content-modeler")}
```

mindctrl:
Added some tests here. @matthewguywright @nickcernis @ChrisWiegman @anthonyburchell

nickcernis (approved):
Looking good! Thanks for adding the tests, @mindctrl.

The “Remove Media” button placement can perhaps be improved, but we can fix with any other things flagged in design reviews later: 

Screenshot 2021-06-30 at 12 36 20

mindctrl (approved):

PR #173, opened 2021-06-23
BH-1073: docs: add architectural decision record for storing taxonomy schemas

mindctrl:


docs/adr/0002-taxonomy-definition-schema.md
```diff
@@ -0,0 +1,58 @@
...
+We will store taxonomy definitions in the `wp_options` table using the `atlas_content_modeler_taxonomies` option name. The definitions will be stored as an associative array, where the array keys are the model's slug. Example:
+```
+[
+	'course' => [
+		'singular' => 'Course', // string(50). Required.
+		'plural' => 'Courses', // string(50). Required.
+		'slug' => 'course', // string(32) - cannot exceed 32 chars, see register_taxonomy(). Required.
+		'hierarchical' => false, // boolean - default false, see register_taxonomy()
+		'types' => [ 'recipe' ], // array - post type slugs separated by comma. Required.
+		'api_visibility' => 'private', // string(10) - 'public' or 'private'. default 'private'. Required.
```
nickcernis, line 24: What does `string(10)` mean in this context?

ChrisWiegman, line 24: This would seem to be a safer storage if we used a bool.

mindctrl: It was a bit arbitrary. The string being the relevant part. The length doesn't really matter, and I could remove/change it. Thoughts?

nickcernis: I think unless we need to impose a hard limit in code we could omit the length from the schema (i.e. just `string`).

mindctrl: Thanks! Removed length in https://github.com/wpengine/atlas-content-modeler/pull/173/commits/9bea953cd73a4414da71a0245604ca590b795df1.

apmatthews: Using a bool here would probably mean updating the model schema to use a bool as well. While this is technically being used as a bool right now, I wonder if it might ever store a full capability as a string ( like `manage_categories` or `manage_options` for example)? I don't know if that level of control would be necessary, but it's worth considering.

ChrisWiegman: I haven't, yet, looked at the model schema but I would definitely agree with that assessment. If there is a reason to use this as a true string I'm more than willing to ignore this. My comment is simply based off context in this PR combined with experience in using strings where a bool would have been more appropriate.

nickcernis: Good points about API consistency and future scope to use other strings. Those seem like good reasons to follow suit here for now.

nickcernis (approved):

docs/adr/0002-taxonomy-definition-schema.md
mindctrl: Thanks for the reviews. Solid thoughts all around. Let's stick with the string for now to unblock all the other work. We can continue discussions in Slack if you all are interested.

PR #174, opened 2021-06-23
[BH-1092] Refactor require usage for better stability

ChrisWiegman:
This standardizes the import process with the following features designed to make a user's site more resilient should a file or other issue happen with our plugin. 

1. Renames `ATLAS_CONTENT_MODELER_DIR` to `ATLAS_CONTENT_MODELER_INCLUDES_DIR` to better reflect our plugin architecture and how we were using the content
2. Standardize imports on `ATLAS_CONTENT_MODELER_INCLUDES_DIR`
3. Replaces all instances of `require` with `include` which, though our plugin will still break, will throw a warning instead of a fatal thereby ensuring the user's site keeps running
4. Removes unneeded `_once` from statements which isn't required and is historically a little better on resources.

### DIFF
--- a/includes/content-registration/custom-post-types-registration.php
+++ b/includes/content-registration/custom-post-types-registration.php
@@ -265,7 +265,7 @@ function generate_custom_post_type_args( array $args ): array {
 
 	$singular = $args['singular'];
 	$plural   = $args['plural'];
-	$icon     = include ATLAS_CONTENT_MODELER_INCLUDES_DIR . 'settings/views/admin-entry-icon.php';
+	$icon     = include_once ATLAS_CONTENT_MODELER_INCLUDES_DIR . 'settings/views/admin-entry-icon.php';
 	$labels   = generate_custom_post_type_labels(
 		[
 			'singular' => $singular,
--- a/includes/content-registration/custom-post-types-registration.php
+++ b/includes/content-registration/custom-post-types-registration.php
@@ -265,7 +265,7 @@ function generate_custom_post_type_args( array $args ): array {
 
 	$singular = $args['singular'];
 	$plural   = $args['plural'];
-	$icon     = include_once ATLAS_CONTENT_MODELER_INCLUDES_DIR . 'settings/views/admin-entry-icon.php';
+	$icon     = include ATLAS_CONTENT_MODELER_INCLUDES_DIR . 'settings/views/admin-entry-icon.php';
 	$labels   = generate_custom_post_type_labels(
 		[
 			'singular' => $singular,
--- a/includes/content-registration/custom-post-types-registration.php
+++ b/includes/content-registration/custom-post-types-registration.php
@@ -265,7 +265,7 @@ function generate_custom_post_type_args( array $args ): array {
 
 	$singular = $args['singular'];
 	$plural   = $args['plural'];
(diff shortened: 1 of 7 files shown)


apmatthews (commented):
I like the consistency here and the improvements you outlined make sense.

There is one more instance of ATLAS_CONTENT_MODELER_DIR in includes/publisher/class-publisher-form-editing-experience.php line 406 that needs updating.

I'm curious what others think about renaming ATLAS_CONTENT_MODELER_DIR to ATLAS_CONTENT_MODELER_INCLUDES_DIR. I have often found it important to have easy access to the true root path of the plugin. While we don't have much outside of the `/includes` directory at the moment, my preference would be to retain ATLAS_CONTENT_MODELER_DIR and manually append `/includes` where necessary. We could also explore the option of keeping both constants or using some kind of helper function.

ChrisWiegman:
Good catch on the missing include update. I'm finding that a search of the plugin folder in VSCode doesn't always get all instances of the search query. I'll be watching more closer for that going forward.

As to the 2nd point, the use of the constant itself, I wrestled with this when I created it and did so based on the architecture patterns we've followed. I can definitely adjust that if we want. This way keeps its usage just a bit easier for anyone who needs it.

includes/content-registration/custom-post-types-registration.php
```diff
@@ -265,7 +265,7 @@ function generate_custom_post_type_args( array $args ): array {
 
 	$singular = $args['singular'];
 	$plural   = $args['plural'];
-	$icon     = require ATLAS_CONTENT_MODELER_DIR . '/includes/settings/views/admin-entry-icon.php';
+	$icon     = include_once ATLAS_CONTENT_MODELER_INCLUDES_DIR . 'settings/views/admin-entry-icon.php';
```
nickcernis, line 268: This is a case where it has to be `include` instead of `include_once`, otherwise we only get the custom admin menu icon for the first CPT (cats in the screenshot below) and not subsequent ones (sloths):

Screenshot 2021-06-23 at 17 43 14

```suggestion
	$icon     = include ATLAS_CONTENT_MODELER_INCLUDES_DIR . 'settings/views/admin-entry-icon.php';
```

includes/publisher/class-publisher-form-editing-experience.php
```diff
@@ -403,7 +403,7 @@ public function should_show_feedback_banner(): bool {
 	 * Runs an `admin_notices` hook.
 	 */
 	public function render_feedback_notice(): void {
-		include_once ATLAS_CONTENT_MODELER_DIR . '/includes/shared-assets/views/banners/atlas-content-modeler-feedback-banner.php';
+		include_once ATLAS_CONTENT_MODELER_INCLUDES_DIR . '/includes/shared-assets/views/banners/atlas-content-modeler-feedback-banner.php';
```
nickcernis, line 406: I think we need to drop the `/includes/` here if `ATLAS_CONTENT_MODELER_INCLUDES_DIR` is going to contain `/includes/`?

```suggestion
		include_once ATLAS_CONTENT_MODELER_INCLUDES_DIR . 'shared-assets/views/banners/atlas-content-modeler-feedback-banner.php';
```

nickcernis (approved):

PR #176, opened 2021-06-23
BH-1053: icon picker using jquery

matthewguywright:


includes/settings/js/src/components/CreateContentModel.jsx
```diff
@@ -194,6 +196,37 @@ export default function CreateContentModel() {
...
 
+					<div
+						className={
+							errors.iconPicker ? "field has-error" : "field"
+						}
+					>
+						<label htmlFor="iconPicker">Model Icon</label>
+						<br />
+						<p className="help">
+							Choose an icon to represent your model.
```
ChrisWiegman, line 207: These should be translatable.

includes/settings/js/src/components/CreateContentModel.jsx
```diff
@@ -194,6 +196,37 @@ export default function CreateContentModel() {
...
+							formRegister={register}
+						/>
+
+						<p className="field-messages">
+							{errors.iconPicker &&
+								errors.iconPicker.type === "required" && (
+									<span className="error">
+										<Icon type="error" />
+										<span role="alert">
+											An icon is required
```
ChrisWiegman, line 223: i18n here too (I'll stop calling these out individually but all strings should be ready for i18n)

includes/settings/js/src/components/IconPicker.jsx
```diff
@@ -0,0 +1,459 @@
...
+	"pets",
+	"privacy",
+	"superhero",
+	"superhero-alt",
+	"edit-page",
+	"fullscreen-alt",
+	"fullscreen-exit-alt",
+	"",
+	"",
+	"",
```
ChrisWiegman, line 311: Is there a reason for the multiple empty options?

Also, is there any easier way to include this array so that it will be easier to maintain as core changes in the future?

ChrisWiegman (changes requested):
The i18n strings need translation. I'm a little concerned about DRY in the icon list but I'm not even sure anything can be done about that.

includes/settings/js/src/components/IconPicker.jsx
matthewguywright: I don't see a programmatic way to include this list. Empty options are removed.

includes/settings/js/src/components/EditModelModal.jsx
```diff
@@ -222,6 +223,44 @@ export function EditModelModal({ model, isOpen, setIsOpen }) {
...
+						</label>
+						<p className="help">
+							{__(
+								"Choose an icon to represent your model.",
+								"atlas-content-modeler"
+							)}
+						</p>
+
+						<IconPicker
+							buttonLabel="Choose Icon"
```
ChrisWiegman, line 242: One more (sorry)

includes/settings/js/src/components/IconPicker.jsx
```diff
@@ -0,0 +1,456 @@
...
+			setIcon(target.val());
+			removePopup();
+		});
+
+		var control = popup.find(".dashicon-picker-control");
+
+		control.html(
+			'<a data-direction="back" href="#">' +
+				'<span class="dashicons dashicons-arrow-left-alt2"></span></a>' +
+				'<input type="text" class="" placeholder="Search" />' +
```
ChrisWiegman, line 385: and another i18n

includes/settings/vendor/dashicons-picker/dashicons-picker.js
```diff
@@ -0,0 +1,439 @@
...
+						.addClass("dashicons-" + title);
+					removePopup();
+				});
+
+				var control = popup.find(".dashicon-picker-control");
+
+				control.html(
+					'<a data-direction="back" href="#">' +
+						'<span class="dashicons dashicons-arrow-left-alt2"></span></a>' +
+						'<input type="text" class="" placeholder="Search" />' +
```
ChrisWiegman, line 382: i18n

ChrisWiegman (dismissed):

ChrisWiegman (dismissed):

nickcernis (commented):
This is working really well for me! Some questions/suggestions for improved UX:

1. Should “Choose Icon” say “Change Icon” when an icon has been picked?
2. Should we add a way to remove an icon to restore the default? (Fine if the answer is “no” or if we just add to backlog for now.)
3. The icon picker modal does not gain keyboard focus. (Tab to “Choose icon”, press space to activate the modal, then try tabbing to an icon to choose it without using the mouse.) Is there a way we can fix this? It means the field is not accessible.
4. Are you comfortable writing e2e tests for the icon picker and sidebar?

I think we should also talk about how the “edit model” modal isn't scaling very well to hold all of these new options we're adding, but happy to shelve that for a separate time. (Perhaps we should discuss reusing the “New Model” page for Edit Model actions with Mason.)

mindctrl:
> Should “Choose Icon” say “Change Icon” when an icon has been picked?

👍  nice touch.

> I think we should also talk about how the “edit model” model isn't scaling very well to hold all of these new options we're adding, but happy to shelve that for a separate time. (Perhaps we should discuss reusing the “New Model” page for Edit Model actions with Mason.)

Agreed, I did mention this to Mason on Monday while demoing our latest work. I'll add it to the backlog so he can start thinking about it.

mindctrl:
@matthewguywright can you bring this up to date with `main`?

mindctrl (changes requested):
We should add some tests here too.

matthewguywright:
> We should add some tests here too.

I will need some assistance in how to go about writing tests for this. @mindctrl

matthewguywright:
@nickcernis @mindctrl I updated based on all the points above. I am going to start working on tests.

mindctrl:
> I will need some assistance in how to go about writing tests for this.

@matthewguywright we should add some kind of e2e test that confirms the icon picker is on the page and can be used. Once a selection is made and saved, confirm it’s there after saving and that it's used in the sidebar.

Also it'd be nice to have a test for the edit modal functionality to confirm it saves there as well.

We could also add some integration test code to confirm that the CSS and JS assets are enqueued when landing on the page in general, though if it's not the e2e test would fail... but it'd be nice to fail earlier on the faster tests and doing an integration test would help that.

mindctrl:
This PR makes the edit modal larger than the screen for me, so I can't see the `h2` title.

Since this field was added between API Identifier and API Visibility, it causes both to go full width when they should be side by side. I think moving the icon field `div` below the API Visibility `div`, instead of in between the two API fields, will mostly fix it.

includes/settings/js/src/components/CreateContentModel.jsx
```diff
@@ -194,6 +196,44 @@ export default function CreateContentModel() {
 						</p>
 					</div>
 
+					<div
+						className={
+							errors.modelIcon ? "field has-error" : "field"
+						}
+					>
+						<label htmlFor="modelIcon">
+							{__("Model Icon", "atlas-content-modeler")}
```
mindctrl, line 205: Let's move this whole field below the API fields. Having so much distance between the Singular Name field and the API Identifier makes it less obvious on smaller screens that the latter is auto-generating as you type in the singular field.

mindctrl (dismissed):

ChrisWiegman (dismissed):
I did find a bug. On a new model I clicked the "choose icon" button but didn't change it. It then wouldn't let me save it as it wanted me to choose one.

Screen Shot 2021-06-25 at 11 34 22 AM

mindctrl:
@ChrisWiegman good catch. I think we can make this field not required since we're supplying a fallback. Thoughts @wpengine/blockheads?

ChrisWiegman:
I think that's the right approach, @matthewguywright.

mindctrl:
These codeclimate recommendations are good, but I'm not refactoring Matthew's work in this PR. If we can make it not "fail" on suggestions that would be great. cc: @ChrisWiegman

includes/content-registration/custom-post-types-registration.php
```diff
@@ -294,7 +294,7 @@ function generate_custom_post_type_args( array $args ): array {
 		'show_in_graphql'       => $args['show_in_graphql'] ?? true,
 		'graphql_single_name'   => $args['graphql_single_name'] ?? camelcase( $singular ),
 		'graphql_plural_name'   => $args['graphql_plural_name'] ?? camelcase( $plural ),
-		'menu_icon'             => $icon,
+		'menu_icon'             => $args['modelIcon'] ?? 'dashicons-admin-post',
```
mindctrl, line 297: The camelCase here doesn't match the snake_case we've used elsewhere.

mindctrl:
Converted to draft until I have a chance to finish converting modelIcon to model_icon where appropriate.

nickcernis:
@mindctrl Is the default icon sticking for you? For me, I see no icon if I don't choose one:

1. Create a model but don't choose an icon (shows a default of the pin icon).
2. The resulting sidebar row has no icon (on creation, and after refresh).
3. The icon appears as `'model_icon' => '',` in the `atlas_content_modeler_post_types` option.

mindctrl:
@nickcernis good catch. Pretty sure that was working, but looks like it's not now. I'll take another look.

mindctrl:
@nickcernis should be fixed in 5b5bf0d7310bf0b5c2a99c819bf6ed9eefdef3f1

nickcernis:
@mindctrl Working for me when the model is first added now, but the default icon disappears from the sidebar on refresh. 😭  (Sorry not to dig in to discover why just yet — happy to circle back in the morning.)

nickcernis (approved):

PR #179, opened 2021-06-24
[BH-1091] Verify plugin zip before deploying

ChrisWiegman:
This PR significantly refactors our CircleCI build in order to allow us to run e2e tests based on the built zip that will then be uploaded for distribution. This will ensure the code were testing is the very code our users will be running and will prevent issues that may arise due to obsolete lines in .zipignore or for any other failure during zip creation.

To get to this point is what took so much refactoring. Here's how it all breaks down:
1. The zip has to be available before the e2e testing. This was a significant departure from previous builds as the way we had been persisting assets would break previously. Now tests do not persist assets and other jobs only persist what they need to. This allows us to build the zip earlier and to run more jobs in parallel which cut off about 2-3 minutes of total build time
2. To sort out which steps were doing what I renamed jobs and re-ordered them to make finding and working with a job a little bit easier. This is the bulk of the refactor

Attached below is both the last build output as well as its associated zip artifact for verification

Screen Shot 2021-06-24 at 11 30 05 AM

[atlas-content-modeler.0.4.1.zip](https://github.com/wpengine/atlas-content-modeler/files/6710442/atlas-content-modeler.0.4.1.zip)

edited to reflect that I had missed php linting in the original build. It has now been refactored into its own job to better reflect failure when it occurs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -314,10 +314,7 @@ jobs:
 
   plugin-test-lint-php:
     working_directory: .
-    docker:
-      - image: php:7.4-apache
-        environment:
-          APP_ENV: test
+    executor: php/default
     steps:
       - attach_workspace:
           at: .
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -461,7 +461,7 @@ workflows:
             tags:
               only: /.*/
       - plugin-deploy:
-          name: "Deploy zip to api (staging) atlas-content-modeler"
+          name: "plugin-deploy-staging"
           slug: atlas-content-modeler
           requires:
             - plugin-build-zip
@@ -479,10 +479,10 @@ workflows:
           auth_url: https://auth-staging.wpengine.io/v1/tokens
           upload_url: https://wp-product-info-staging.wpesvc.net/v1/plugins
       - plugin-deploy:
-          name: "Deploy zip to api (production) atlas-content-modeler"
+          name: "plugin-deploy-production"
           slug: atlas-content-modeler
           requires:
-            - "Deploy zip to api (staging) atlas-content-modeler"
+            - "plugin-deploy-staging"
           filters:
             branches:
               ignore: /.*/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -461,7 +461,7 @@ workflows:
       - plugin-build-json:
           slug: atlas-content-modeler
           requires:
-            - plugin-test-unit
+            - plugin-build-composer
           # Run this job on every commit/PR to make sure it's in working order prior to deploying
           filters:
             tags:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -209,6 +209,7 @@ jobs:
       - run:
           name: Run Code Sniffer sniffs
           command: vendor/bin/phpcs
+          working_directory: atlas-content-modeler
 
   plugin-test-unit:
     docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -191,6 +191,25 @@ jobs:
             npm run lint
           working_directory: <<parameters.slug>>
 
+  plugin-test-lint-php:
+    working_directory: .
+    docker:
+      - image: php:7.4-apache
+        environment:
+          APP_ENV: test
+    steps:
+      - attach_workspace:
+          at: .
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -396,102 +269,240 @@ jobs:
...
+      - run:
+          name: Install system packages
+          command: apt-get update && apt-get -y install git
+      - run:
+          name: Install PHP extensions
+          command: docker-php-ext-install pdo
+      - run:
+          name: Install Composer
+          command: |
+            php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
```
markkelnar, line 333: Why are you install composer like this instead of the php/install-composer orb one?

ChrisWiegman: Honest answer... I'm learning Circle as I'm doing this. This snippet is actually something I found in one of their forums

.circleci/config.yml
```diff
@@ -183,55 +105,7 @@ jobs:
           paths:
             - build
 
-  plugin-deploy:
```
markkelnar, line 186: Moving sections of code like this, for what appears to be just sorting of the code and not much change in behavior, makes line-by-line comparison difficult.

.circleci/config.yml
```diff
@@ -183,55 +105,7 @@ jobs:
...
-          user: WPE_LDAP_USER
-          pass: WPE_LDAP_PASS
-          url: <<parameters.auth_url>>
-      - wp-product-orb/post-zip:
-          url: <<parameters.upload_url>>/<<parameters.slug>>
-          zip: build/<<parameters.slug>>.$BUILD_VERSION.zip
-          json: build/<<parameters.slug>>.$BUILD_VERSION.json
-          version: $BUILD_VERSION
-
-  plugin-npm-build:
```
markkelnar, line 211: For example, to the above comment, like where did this go? Name of the job changed? Or did we drop the functionality of what has happening here?  I think its now 'plugin-build-npm' but not sure.

.circleci/config.yml
markkelnar: Fair enough.  The php orb one comes with some built in caching for speed and reuse. It may not solve exactly what you want to do, but there's a good change it will.

markkelnar: I got in the habit of first looking for an official orb from circleci that does functionality I want to see if there's already something avail.

.circleci/config.yml
ChrisWiegman: I do not disagree at all. Normally it is a type of refactor I wouldn't want to do. In this case I struggled to understand both the flow and structure of the existing pipeline as I learned how CircleCI handles these in general. In refactoring the code to such an order I was finally able to make more sense of it. I left the changes as, while they're inconvenient here, I do believe they'll be easier for folks in the future

.circleci/config.yml
ChrisWiegman: I've updated this to use the install script and realized that you had, in fact, been running phpcs. As the composer job was something that was working before I hadn't found a need to change that and missed that step. I've kept the composer install as is but moved the lint to a different job so that a failure in php linting will show as a fail in php linting instead of the composer build

.circleci/config.yml
markkelnar: tomayto, tomahto

Still makes this review difficult to catch any bugs. So, gonna have to blindly trust it just works.

.circleci/config.yml
```diff
@@ -338,8 +206,7 @@ jobs:
       - run:
           name: Activate atlas-content-modeler plugin
           command: |
-            ln -s ~/project/atlas-content-modeler /tmp/wordpress/wp-content/plugins/atlas-content-modeler
-            ./wp-cli.phar plugin activate --path=/tmp/wordpress atlas-content-modeler
+            ./wp-cli.phar plugin install --activate --path=/tmp/wordpress build/$(cat build/file.txt)
```
nickcernis, line 209: 👍 Nice idea to grab the zip artifact to e2e test like this.

nickcernis (approved):
This looks good to me, @ChrisWiegman.

I checked the CI logs and verified the zip is working. Thanks for making this easy to test and for giving us more confidence that zips we tag will be good.

mindctrl (approved):

PR #185, opened 2021-06-29
[BH-1054] Add CodeClimate config

ChrisWiegman:
This adds a very basic codeclimate config along with 5 of their plugins:

* phpmd
* eslint (probably isn't needed but doesn't hurt anything)
* duplication 
* sonar-php
* fixme

(see https://docs.codeclimate.com/docs/list-of-engines)

Note this is configured to get us started. The complexity checks do require a different way of looking at code, from experience with them. We can tune further as we learn what works best for us. In addition, I've excluded jsx for now as the logic for the complexity checks simply throws too many false positives with it.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,32 @@
+version: "2"         # required to adjust maintainability checks
+checks:
+  argument-count:
+    config:
+      threshold: 4
+  complex-logic:
+    config:
+      threshold: 4
+  file-lines:
+    config:
+      threshold: 250
+  method-complexity:
+    config:
+      threshold: 5
+  method-count:
+    config:
+      threshold: 20
+  method-lines:
+    config:
+      threshold: 25
+  nested-control-flow:
+    config:
+      threshold: 4
+  return-statements:
+    config:
+      threshold: 4
+  similar-code:
+    config:
+      threshold: # language-specific defaults. an override will affect all languages.
+  identical-code:
+    config:
+      threshold: # language-specific defaults. an override will affect all languages.
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -8,7 +8,7 @@ checks:
       threshold: 4
   file-lines:
     config:
-      threshold: 250
+      threshold: 500
   method-complexity:
     config:
       threshold: 5
@@ -30,6 +30,8 @@ checks:
   identical-code:
     config:
       threshold: # language-specific defaults. an override will affect all languages.
+  exclude_patterns:
+    - "*.jsx"
   plugins:
     eslint:
       enabled: true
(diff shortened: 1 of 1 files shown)


.codeclimate.yml
```diff
@@ -0,0 +1,41 @@
...
+      threshold: 4
+  similar-code:
+    config:
+      threshold: # language-specific defaults. an override will affect all languages.
+  identical-code:
+    config:
+      threshold: # language-specific defaults. an override will affect all languages.
+  plugins:
+    rubocop:
+      enabled: true
```
mindctrl, line 35: Do we need `rubocop`? CodeClimate says it's for Ruby code, which we don't have here.

ChrisWiegman: Ha, nope. Got a little carried away with that one. Removed.

anthonyburchell (dismissed):
As much as I could pick up from the docs, this looks right to me. :) 

Excited to see the impact this makes on our processes. Thanks for kicking this off @ChrisWiegman !

.codeclimate.yml
```diff
@@ -0,0 +1,39 @@
...
+      threshold: 4
+  similar-code:
+    config:
+      threshold: # language-specific defaults. an override will affect all languages.
+  identical-code:
+    config:
+      threshold: # language-specific defaults. an override will affect all languages.
+  plugins:
+    eslint:
+      enabled: true
```
nickcernis, line 35: Hey, @ChrisWiegman. I'm new to Code Climate so forgive my stupid questions:

1. What is eslint doing in Code Climate that it's not doing when we run it in Circle? (Is it the complexity analysis?)

2. Do we need to extend it to cover JSX files?

> By default, the ESLint engine only analyzes .js files. If you want ESLint to analyze your JSX code and ES6 files, you'll need to specify this in your .codeclimate.yml, as shown in the example above.

https://docs.codeclimate.com/docs/eslint#extensions

nickcernis (dismissed):
Looking good. I just had a couple of small questions above about eslint, but they don't block merge.

.codeclimate.yml
```diff
@@ -0,0 +1,39 @@
...
+      threshold: 250
+  method-complexity:
+    config:
+      threshold: 5
+  method-count:
+    config:
+      threshold: 20
+  method-lines:
+    config:
+      threshold: 25
```
nickcernis, line 20: This `method-lines` length threshold looks like it's tuned for Ruby code to me. JS methods are often longer than 25 lines once Prettier gets involved in formatting JSX. I'm happy to see how it goes if you think it's a good default to start with, though!

ChrisWiegman: Yeah, these are the ones I'm torn on. PHP, JS and Ruby often want different values. We've gone has high as 20 for complexity in Go. I like starting these smaller and then we can see what we need to tune later.

mindctrl (approved):
LGTM as a starting point. We can tweak the values as necessary.

nickcernis (approved):

PR #186, opened 2021-06-29
[BH-1109] fix: handle lowercased model labels in graphiQL query generator

apmatthews:
## Description

- Fixes invalid GraphiQL queries generated from models that use lowercased names
- Fixes invalid GraphiQL queries generated on models that have no fields
- Adapts to new WPGraphQL convention of prepending the singular name with `all` to generate the plural field name when singular and plural names are identical

We'll want to circle back to this before we start adding pre-generated queries in too many other places, but this should patch things for now. I looked into using GraphQL introspection queries to look up the exact format of the schema as it is registered in WPGraphQL and while definitely possible, it's going to be a much bigger task.

https://wpengine.atlassian.net/browse/BH-1109

## Testing

To test models with lowercased names and no fields:

1. Create a model using lowercased 'cat' and 'cats' as the singular and plural names, respectively (case sensitive)
2. In the model's "..." options menu, click "Open in GraphiQL".
3. See that there are no linting errors and the query runs successfully.

To test models with same singular and plural names:

1. Make sure you're running at least WPGraphQL v1.5.0.
2. Create a model using lowercased 'deer' and 'deer' as the singular and plural names, respectively
3. In the model's "..." options menu, click "Open in GraphiQL".
4. See that there are no linting errors, the query includes `allDeer`, and the query runs successfully.

### DIFF
--- a/includes/settings/js/src/utils.js
+++ b/includes/settings/js/src/utils.js
@@ -101,8 +101,11 @@ export const maybeCloseDropdown = (setDropdownOpen, timer) => {
  * @return {string} The GraphiQL URL with query param prefilled.
  */
 export const getGraphiQLLink = (modelData) => {
-	const modelSingular = modelData.singular.replace(/\s/g, "");
-	const fragmentName = `${modelSingular}Fields`;
+	const graphQLType = toValidApiId(modelData.singular).replace(
+		/^[a-z]/g,
+		(match) => match.toUpperCase() // GraphQL's "Types" are all capitalized
+	);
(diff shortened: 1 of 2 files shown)


nickcernis (dismissed):
This is working well! Thanks for taking care of both cases, and for the upstream fix in WPGraphQL. 🏆

Screenshot 2021-07-07 at 10 19 31
Screenshot 2021-07-07 at 10 18 35

nickcernis (approved):

ChrisWiegman (approved):

PR #183, opened 2021-06-29
BH-1075 Allow creation of taxonomies in the developer app

nickcernis:
- Adds a new taxonomies view at `/wp-admin/?page=atlas-content-modeler&view=taxonomies`.
- Adds a “View Taxonomies” button on the main model listings page (for now; this may move later).

Screenshot 2021-06-30 at 11 43 32

Screenshot 2021-06-30 at 11 43 21

### To test
1. Visit the developer app at `/wp-admin/admin.php?page=atlas-content-modeler`.
2. Click “View Taxonomies”.
3. Try adding taxonomies. You should see them in the list to the right of the taxonomies form.
4. Run `wp option get atlas_content_modeler_taxonomies` or view that row in the WP options table to see the result of your submission as data.
 
### Notes
- We plan to improve the listing of taxonomies in [BH-1078](https://wpengine.atlassian.net/browse/BH-1078) (“List Taxonomies”). I added the listing area to the right of the form as a placeholder for now just to wire up the data.
- Perhaps we could re-use the existing form in [BH-1076](https://wpengine.atlassian.net/browse/BH-1076) (“Edit Taxonomy”)? We could hide the list view in edit mode and present only the form with submit/cancel buttons.

PR #187, opened 2021-07-01
BH-1078: Add Taxonomy List

nickcernis:
- Implements the taxonomy table on the taxonomies developer app view.
- Adds the taxonomy options button and drop down.
- Adds “Edit” and “Delete” options to the drop down. 

### Notes
- Edit and Delete options are wired up to work but will show a TODO alert for now. (This targets the 'taxonomies' branch, so we won't release it until this is fixed in separate PRs.)
- We decided in Slack to remove the Bulk Action option for now.
- I cleared the column heading changes with Mason. (The [Figma mockup](https://www.figma.com/file/6b3kxG7r2OTv2A024CQiUU/Taxonomy) has different headings.)

### To test
1. Click “View Taxonomies from the main model page or visit `/wp-admin/?page=atlas-content-modeler&view=taxonomies` or 
2. Add one or more taxonomies, which should appear in the table.
3. Try the “Edit” or “Delete” actions, which should present you with an alert.

### Before
Screenshot 2021-07-01 at 18 51 23

### After
Screenshot 2021-07-01 at 18 04 52

### DIFF
--- a/includes/settings/js/src/components/Taxonomies.jsx
+++ b/includes/settings/js/src/components/Taxonomies.jsx
@@ -6,6 +6,7 @@ import { useForm } from "react-hook-form";
 import Icon from "../../../../components/icons";
 import { useApiIdGenerator } from "./fields/useApiIdGenerator";
 import { showSuccess } from "../toasts";
+import TaxonomiesTable from "./TaxonomiesTable";
 
 const { apiFetch } = wp;
 
@@ -99,7 +100,7 @@ export default function Taxonomies() {
 	}
 
 	return (
-		<div className="app-card">
+		<div className="app-card taxonomies-view">
 			<section className="heading flex-wrap d-flex flex-column d-sm-flex flex-sm-row">
 				<h2>{__("Taxonomies", "atlas-content-modeler")}</h2>
 				<button
@@ -491,13 +492,8 @@ export default function Taxonomies() {
 							</button>
 						</form>
 					</div>
-					<div className="taxonomy-list col-xs-10 col-lg-6 order-0 order-lg-1">
-						{/* TODO: Display taxonomies in a table here. */}
-						{Object.values(taxonomies).map((taxonomy) => {
-							return (
-								<p key={taxonomy?.slug}>{taxonomy?.plural}</p>
-							);
-						})}
+					<div className="taxonomy-list col-xs-10 col-lg-7 order-0 order-lg-1">
+						<TaxonomiesTable taxonomies={taxonomies} />
 					</div>
 				</div>
 			</section>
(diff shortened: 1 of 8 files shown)


ChrisWiegman (commented):
Can you please better describe the testing you're doing on this? It looks like there are acceptance tests. Is that all?

I also don't think we should require a model be selected. This might make creating taxonomies easier in that they can be assigned later and opens us up to a better/v2 workflow of allowing a user to pick a taxonomy from the model screen. (screenshot of where I'm talking about is below)

Screen Shot 2021-07-01 at 12 25 16 PM

nickcernis:
Hi @ChrisWiegman! Thanks for testing it out. I think you may have missed https://github.com/wpengine/atlas-content-modeler/pull/183, which implemented the form, submission logic, and associated tests you mention. This PR just adds a table where a bare list of models used to display.

I updated the first comment above with a before and after screenshot in case that helps.

> Can you please better describe the testing you're doing on this?

This PR only changes output of the model list from showing a list of plural names to showing a table. The change to the e2e test I made reflects that this was the only testable change.

> It looks like there are acceptance tests. Is that all?

Tests for the dependent REST endpoints and the e2e tests for the form functionality were added in https://github.com/wpengine/atlas-content-modeler/pull/183. We should of course also add tests for the upcoming taxonomy edit/delete functionality in the PRs for those features.

> I also don't think we should require a model be selected. 

This was part of the acceptance criteria for https://github.com/wpengine/atlas-content-modeler/pull/183 / [https://wpengine.atlassian.net/browse/BH-1075](BH-1075):

> User can associate the taxonomy with one or more models during creation.

(One or more, not zero or more.) Happy to chat about the implications and remove that limitation in a separate PR if we need to. I wonder if there's a risk of someone skipping the model list and wondering why their taxonomy isn't in use anywhere, though? We need to put some thought into this before undoing it.

nickcernis:
Updated with an empty table state and test for the empty state.

I added a [Jira story](https://wpengine.atlassian.net/browse/BH-1111) to make a PR template to prompt to include context about tests in future PRs. Thanks for the feedback, @ChrisWiegman. ❤️

ChrisWiegman (approved):
Thank you for the answers Nick. Indeed I had not reviewed #183 so the context of this makes much more sense now. Approving.

apmatthews (approved):
LGTM! Down the road we'll probably want to extract a more generic version of the dropdown component that can be shared between models and taxonomies, but I actually like having two fully coded use cases as a starting point for that kind of extraction. It makes it easier to see what props a generic version might need and where different responsibilities should lie.

mindctrl (approved):

nickcernis:
Thanks for the reviews, all!

> Down the road we'll probably want to extract a more generic version of the dropdown component that can be shared between models and taxonomies…

💯 Completely agree. We should review the places we're using a modal and abstract that too. I'll add backlog items for both.

PR #190, opened 2021-07-06
BH-1076: edit taxonomies

nickcernis:
Implements taxonomy editing for [BH-1076](https://wpengine.atlassian.net/browse/BH-1076)

## Testing

- Includes e2e tests for taxonomy editing.
- Taxonomy update REST integration tests already exist. 

To test manually:

1. Visit the developer app home screen at `/wp-admin/?page=atlas-content-modeler`
2. Click “View Taxonomies”.
3. Create a taxonomy.
4. Edit your taxonomy by choosing Edit in the taxonomy options dropdown.

## Notes
- The acceptance criteria say, “If user tries to save a taxonomy with a duplicate label or plural label, an error is displayed and the data is not saved”. I have not implemented this because we only need to prevent taxonomy slug conflicts (which we already do during taxonomy creation), not singular/plural name conflicts. A user may want two taxonomies with the same name but different slugs, so that different models could be categorised with unique terms under a shared name. (I might want two taxonomies named “Category” with 'car-category' and 'house-category' slugs, so that I can have car category terms and house category terms instead of having to share category terms between houses and cars.)
- I made the “hierarchical” field uneditable (“Terms can have parent terms” can only be set during taxonomy creation). Changing this after a category is created could have unexpected side-effects for already-entered data.
- I broke the form into its own component, but Code Climate complains about the length (483 lines) and code repetition in `TaxonomiesForm.jsx`. We could improve this, but I think it's better done as a separate PR to extract form components and refactor admin form handling across the setting app. This is already captured in part as [BH-1110](https://wpengine.atlassian.net/browse/BH-1110).

## Screenshots

Screenshot 2021-07-19 at 13 33 46

Screenshot 2021-07-19 at 13 34 22

## Documentation Changes

n/a

## Dependant PRs

n/a


mindctrl:
This is working nicely!

A thought: editing a taxonomy doesn't update the URL. If you click "View Taxonomies", edit an existing taxonomy, click the back button in the browser without making changes, you return to the models listing and not the taxonomies listing.

We chose not to update the URL when editing _models_, because we currently do that in a modal and at the time we agreed it didn't feel right to update history to include the modal view. But here we're not using a modal (yay!). Should we update the URL to include the id/slug of the taxonomy being edited? This would retain typical back/forward browser button behavior, and allow a browser refresh to return back to the edit taxonomy screen instead of the listing screen (similar to how the model view works).

Thoughts?

apmatthews (commented):
Everything is working as expected with the one exception noted in the comments.

At some point we may want to revisit validation on the PHP side, though I don't know if it should be a blocker here since that work was handled earlier. I think we need to make sure we're enforcing the restrictions that only exist as front-end validation or HTML attributes. Some examples include:

1. At least one model association is required.
2. The "hierarchical" setting cannot be changed for existing taxonomies.
3. API identifiers cannot be changed for existing taxonomies.

The first two restrictions could be bypassed by modifying the DOM or by sending your own PUT request to the API. The third restriction technically can't be accomplished through our endpoints, but someone tinkering with the taxonomy slug in a PUT request could end up creating a new taxonomy rather than updating an existing one. Not terrible, but I listed it anyway for completeness.

tests/acceptance/EditTaxonomyCest.php
```diff
@@ -0,0 +1,37 @@
...
+		$I->wait(1);
+		$I->amOnTaxonomyListingsPage();
+		$I->wait(1);
+	}
+
+	public function i_can_edit_a_taxonomy(AcceptanceTester $I)
+	{
+		$I->click('.action button.options');
+		$I->see('Edit', '.dropdown-content');
+        $I->click('Edit', '.action .dropdown-content');
```
mindctrl, line 21:
```suggestion
		$I->click('Edit', '.action .dropdown-content');
```

spaces > tabs

nickcernis: Thank you! Fixed in https://github.com/wpengine/atlas-content-modeler/pull/190/commits/a06686f3457e75b4136d3df10590bd8e82c76bb0.

nickcernis: Perhaps we should also have phpcs check the test files to save us from commenting on formatting during review.

includes/settings/js/src/components/TaxonomiesForm.jsx
```diff
@@ -0,0 +1,482 @@
...
+			typeof data.types === "boolean" ||
+			(typeof data.types === "object" && data.types.length < 1)
+		) {
+			setError("types", {
+				type: "noModelSet",
+			});
+			return false;
+		}
+
+		if (editingTaxonomy && !isDirty) {
```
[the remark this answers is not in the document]
nickcernis: Thanks, @apmatthews. Great suggestion.

Should be addressed in:
- https://github.com/wpengine/atlas-content-modeler/pull/190/commits/a705b4463af4ee0463c0afedb74b7baf33bdc86b
- https://github.com/wpengine/atlas-content-modeler/pull/190/commits/a06686f3457e75b4136d3df10590bd8e82c76bb0

nickcernis:
Just refactoring to address @mindctrl's suggestion from https://github.com/wpengine/atlas-content-modeler/pull/190#issuecomment-882821508.

nickcernis:
> A thought: editing a taxonomy doesn't update the URL. If you click "View Taxonomies", edit an existing taxonomy, click the back button in the browser without making changes, you return to the models listing and not the taxonomies listing.

Good suggestion, @mindctrl. I made that change in https://github.com/wpengine/atlas-content-modeler/pull/190/commits/938d9604dbbf0ef9ea6b94dd43f5f172d9069c89 and added a test in https://github.com/wpengine/atlas-content-modeler/pull/190/commits/9b8844023dede9ca50e61a0711dbefaf64fcb0f2.

This is ready for another look.

It might also unblock the menu work once merged, @apmatthews. I'll file a separate story about the REST issues you kindly flagged.

apmatthews (approved):

mindctrl (approved):

PR #199, opened 2021-07-21
BH-1103: number field type validation fixes

matthewguywright:
## Description

### DIFF
--- a/includes/settings/js/src/components/CreateContentModel.jsx
+++ b/includes/settings/js/src/components/CreateContentModel.jsx
@@ -198,12 +198,12 @@ export default function CreateContentModel() {
 
 					<div className={errors.slug ? "field has-error" : "field"}>
 						<label htmlFor="slug">
-							{__("API Identifier", "atlas-content-modeler")}
+							{__("Model ID", "atlas-content-modeler")}
 						</label>
 						<br />
 						<p className="help">
 							{__(
-								"Auto-generated from the singular name and used for API requests.",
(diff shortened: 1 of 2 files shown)


ChrisWiegman (commented):
It's highly possible I'm missing something but, using the steps laid out in Jira, I'm still receiving the error when inputing a decimal...

Screen Shot 2021-07-21 at 12 04 23 PM

ChrisWiegman (dismissed):
This looks much better. LGTM

Screen Shot 2021-07-21 at 1 06 10 PM

ChrisWiegman (dismissed):

apmatthews (changes requested):
Working well!

I saw the discussion in Slack about catching negative numbers when the min is set to 0. I think we can address that here by just changing the conditionals that check for the existence of min/max/step settings. It looks like these values are stored as numbers in the field settings, not strings. Setting min to 0 makes `field?minValue` evaluate to `false`, which results in the `min` attribute being left off of the input.

I would also add a couple more Jest test cases to cover this new behavior. The two important ones that come to mind are:
1. It renders a proper number field when min/max/step are set (using zero for one these values would be a good way to test for that negative bug indirectly).
2. It renders a default step of 1 for number fields with a type of integer.

mindctrl:
I see an issue with the validation error message, where the decimal is dropped.

Decimal field example:
min value: -12.5
max value: 112.5
step value: 2.5

Enter a number lower than `-12.5` and see the error message says `Minimum value is -12.` ... it's dropping the decimal on the text.

image

PR #200, opened 2021-07-21
[BH-1122] Update "API Identifer" to "Model ID" with associated description

ChrisWiegman:
## Description

Changes “API Identifier” to “Model ID” and updates the associated description.

Screen Shot 2021-07-21 at 12 45 27 PM

Screen Shot 2021-07-21 at 12 45 52 PM

https://wpengine.atlassian.net/browse/BH-1122

This follows the first approach outlined in the Jira ticket above in order to better represent the field and its intended use without confusion in GraphQL or other interfaces.

### DIFF
--- a/includes/rest-api/rest-api-endpoint-registration.php
+++ b/includes/rest-api/rest-api-endpoint-registration.php
@@ -509,7 +509,7 @@ function create_model( string $post_type_slug, array $args ) {
 	if ( empty( $post_type_slug ) ) {
 		return new WP_Error(
 			'atlas_content_modeler_invalid_id',
-			__( 'Please provide a valid API Identifier.', 'atlas-content-modeler' ),
+			__( 'Please provide a valid Model ID.', 'atlas-content-modeler' ),
 			[ 'status' => 400 ]
 		);
 	}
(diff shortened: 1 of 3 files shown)


apmatthews (changes requested):
I think we want to keep "API Identifier" anywhere that it's used to describe the slug of a user-defined "field" (in `Form.jsx` for example). It's accurately worded in that context since we pass the value to REST and WPGraphQL.

nickcernis:
> I think we want to keep "API Identifier" anywhere that it's used to describe the slug of a user-defined "field"

Agreed, it only needs to change for the new model and edit model forms.

ChrisWiegman:
Thanks. This is good to know as I can't deny I don't understand where these all are actually used in context. I've redone the PR with only the changes to the two forms.

nickcernis (dismissed):
Looks great! Thanks for grabbing this one, @ChrisWiegman.

mindctrl (dismissed):

apmatthews:
A couple of instances we do want changed were reverted. See the `create_model()` function error messages.

Screen Shot 2021-07-22 at 9 34 49 AM

ChrisWiegman:
Good catch Andrew. I had manually changed the field but hadn't reproduced an error.
Screen Shot 2021-07-22 at 9 48 14 AM

apmatthews (approved):

mindctrl (approved):

nickcernis (approved):

PR #201, opened 2021-07-21
[BH-1079, BH-1080] Reflect taxonomy updates in admin sidebar

apmatthews:
## Description

Updates the admin sidebar in response to any changes in taxonomies. I elected to go this route rather than mirroring the approach we use for models because of the nested updates that need to happen on changes to a taxonomy's "types" setting. It seemed cleaner to do the necessary diffing and updating all in one place.

https://wpengine.atlassian.net/browse/BH-1079
https://wpengine.atlassian.net/browse/BH-1080

One issue I ran into is deciding what to do with taxonomies that have non-existent types assigned to them. For example:

1. Create Restaurant and Recipe models.
2. Create a Cuisines taxonomy and assign it to both Restaurants and Recipes.
3. Delete the Recipe model.

In this case, we still have 'recipe' defined as a type on Cuisines and it still shows up in the taxonomy list table. I've worked around it in this PR, but it feels like something that warrants more discussion. ~~I'll create a ticket in JIRA with some thoughts to kick it off.~~ Ticket is here: https://wpengine.atlassian.net/browse/BH-1148

## Testing

I've included a set of acceptance tests that cover all taxonomy actions. I also included a couple of tests to cover sidebar updates for model creation/deletion.

### DIFF
--- a/includes/settings/js/src/components/Taxonomies.jsx
+++ b/includes/settings/js/src/components/Taxonomies.jsx
@@ -1,4 +1,4 @@
-import React, { useContext } from "react";
+import React, { useContext, useEffect, useRef } from "react";
 import { useHistory } from "react-router-dom";
 import { __, sprintf } from "@wordpress/i18n";
 import { useLocationSearch } from "../utils";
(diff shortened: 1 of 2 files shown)


includes/settings/js/src/components/Taxonomies.jsx
```diff
@@ -13,6 +13,79 @@ export default function Taxonomies() {
...
+
+				// Give the list item an updated edit link.
+				listItem.innerHTML = `<a href="${href}">${plural}</a>`;
+				// Always show taxonomies that are enabled in case we hid them on removal.
+				listItem.style.display = "list-item";
+
+				// If the taxonomy link didn't exist already, we should insert it.
+				if (!taxLink) {
+					const postTypeSubMenu = document.querySelector(
+						`#menu-posts-${type} .wp-submenu`
```
nickcernis, line 75: I think `type` needs to be lowercase here, for example with this or similar:

```suggestion
						`#menu-posts-${type.toLowerCase()} .wp-submenu`
```

Otherwise a new taxonomy won't get added to a model with a `customCamelCaseApiID`, because the WP menu uses a lowercase `#menu-posts-customcamelcaseapiid`.

(I think we might also have a bug unrelated to this PR where taxonomies aren't being registered correctly for those camelCase models; I don't see the taxonomy next to the model by default after I add the taxonomy and refresh the page, for example.)

nickcernis:
Looks good, @apmatthews! I like the approach with `useEffect` here. Just one edge case as noted above for mixed case model IDs.

includes/settings/js/src/components/Taxonomies.jsx
apmatthews: Nice catch! My head is always spinning trying to keep up with what format our strings need to be in for WordPress, REST, WPGraphQL, HTML output, JS, etc. 🙃 

I looked into the unrelated bug you mentioned and I think we have an opportunity to simplify at least one of those cases. WordPress uses [`sanitize_key()`](https://developer.wordpress.org/reference/functions/sanitize_key/) to format our Model ID when it goes through `register_post_type()`. From there on, it's always lowercased, which is why taxonomies aren't getting wired up correctly for models with camel-cased IDs. As far as I can tell, the Model ID is only used for the WP post type slug, so maybe we should match the `sanitize_key()` behavior in the Model ID generator? I could explore that in a separate PR, but we will have to consider how backwards compatible we need it to be.

Thoughts?

nickcernis: I like the idea of making it lowercase everywhere for the model, effectively treating the model ID santization and other API ID sanitization separately. Perhaps we could lowercase here for now and follow up with a separate PR for the model ID handling as you suggest?

apmatthews:
Thanks for the feedback @nickcernis! I've updated with the lowercasing you suggested for now and opened https://wpengine.atlassian.net/browse/BH-1147 as a more permanent solution.

nickcernis (approved):

mindctrl (approved):

PR #202, opened 2021-07-22
BH-1103: Number field validation reports inaccurate errors

matthewguywright:
## Description


mindctrl:
Linking to comment on the previous PR: https://github.com/wpengine/atlas-content-modeler/pull/199#issuecomment-884882727

ChrisWiegman (dismissed):

anthonyburchell:
I noticed that you are able to use `.0` in integer fields. 

to reproduce, create a new field set to integer and enter `5.0` for the value. This is the same as `5` technically, but I wonder if the decimal will trip folks up on the front end.

anthonyburchell:
Could we potentially do an `onChange` to just not allow decimals to be entered in integer number fields?

mindctrl:
@matthewguywright nice fix on the decimals getting cut off on the error message.

Something @apmatthews said yesterday is worth considering... https://github.com/wpengine/atlas-content-modeler/pull/199#discussion_r674339038

Let me show an example:
Number field, configured as Integer, advanced settings as follows: min: -5, max: 25, step: 5

On publisher side, if someone enters "2" in this field, the error message says "Step value is 5", which is confusing and helpful to understand what type of values can be entered.

Screenshot
image

Thoughts on how to address this in a way that's easy for users to understand? We can do this in another PR if we want more time to think it through.

mindctrl:
@matthewguywright this seems to be working well now. Can you add some e2e tests that confirm the desired behavior?

includes/publisher/js/src/components/Field.jsx
```diff
@@ -43,6 +43,8 @@ export default function Field(props) {
 					),
 					event.target.step.toString()
 				);
+			} else if (event.target.validity.customError) {
+				error = "The input is invalid.";
```
ChrisWiegman, line 47: This needs to be translatable.

ChrisWiegman (changes requested):
This seems to work for me. Only issue is a string that is missing i18n.

includes/publisher/js/src/components/Field.jsx
matthewguywright: Updated

ChrisWiegman (dismissed):

tests/acceptance/PublishModelCest.php
```diff
@@ -2,15 +2,86 @@
...
+		$i->wait(2);
+
+		$i->see('This field is required');
+		$i->wait(1);
+
+		$i->seeInField('atlas-content-modeler[goose][integer]', '');
+		$i->seeInField('atlas-content-modeler[goose][decimal]', '');
+	}
+
+	public function error_is_triggered_for_invalid_value(\AcceptanceTester $i)
```
mindctrl, line 49: @matthewguywright let's rename these test methods to something more descriptive and related to the number field things they're testing. Example: `i_see_submission_errors_in_number_fields_when_input_is_invalid_for_the_number_type` or something along those lines.

includes/publisher/js/src/components/Field.jsx
```diff
@@ -24,6 +24,30 @@ export default function Field(props) {
...
+			} else if (event.target.validity.stepMismatch) {
+				error = sprintf(
+					__(
+						"Value must be a multiple of %s.",
+						"atlas-content-modeler"
+					),
+					event.target.step.toString()
+				);
+			} else if (event.target.validity.customError) {
+				error = __("The input is invalid.", "atlas-content-modeler");
```
mindctrl, line 47: Can we make this error message clearer? Can we tell the user why it's invalid?

mindctrl (changes requested):

tests/acceptance/PublishModelCest.php
```diff
@@ -2,15 +2,52 @@
...
 {
-	public function i_can_publish_a_model_entry(AcceptanceTester $i)
+	public function _before(\AcceptanceTester $i)
 	{
 		$i->maximizeWindow();
 
 		// First we create a model with fields.
 		$i->loginAsAdmin();
-		$i->haveContentModel('goose', 'geese', 'Geese go honk');
+		$i->haveContentModel('moose', 'mooses', 'Mooses go frmh');
```
mindctrl, line 11: It'd be nice to not change these tests from goose to moose. I know that came about due to local test failures, where the content models weren't being wiped out between tests, but I think we are addressing that another way.

matthewguywright: Updated

mindctrl:
@matthewguywright if you bring this up to date with `main` the failing unit tests should pass.

anthonyburchell (dismissed):
Awesome stuff! Great workarounds to the edge cases! LGTM 👍

tests/acceptance/PublishModelCest.php
```diff
@@ -2,15 +2,52 @@
...
 {
-	public function i_can_publish_a_model_entry(AcceptanceTester $i)
+	public function _before(\AcceptanceTester $i)
 	{
 		$i->maximizeWindow();
 
 		// First we create a model with fields.
 		$i->loginAsAdmin();
-		$i->haveContentModel('goose', 'geese', 'Geese go honk');
+		$i->haveContentModel('goose', 'gooses', 'gooses go frmh');
```
mindctrl, line 11:
```suggestion
		$i->haveContentModel('goose', 'geese', 'Geese go honk');
```

...to get it back to what it was before... gooses? :laughing:

tests/acceptance/PublishModelCest.php
```diff
@@ -2,15 +2,52 @@
...
+		$i->wait(1);
+
+		$i->fillField(['name' => 'atlas-content-modeler[goose][integer]'], '');
+		$i->fillField(['name' => 'atlas-content-modeler[goose][decimal]'], '');
+		$i->scrollTo('#submitdiv');
+
+		$i->click('Publish', '#publishing-action');
+		$i->wait(2);
+
+		$i->see('This field is required');
```
mindctrl, line 42: @matthewguywright I wonder if we should test a scenario where the field is not required? The Jira issue described a case where the "This field is required" message showed when it should not. A test that confirms this would catch any regressions for that specific bug.

matthewguywright: Haha. Updated.

anthonyburchell (approved):
Nice callout on that test case @mindctrl ! Looks good to me! :D

mindctrl (approved):

PR #207, opened 2021-07-26
[BH-1148] Remove ACM taxonomy associations when deleting a content model

apmatthews:
## Description

- The `delete_model()` function has been refactored a bit for this work. It didn't appear to be used anywhere, so I thought it made sense to update and align it with how we use `create_model()` and `update_model()`. A side effect of this refactoring is improved error messaging, which allowed for better error handling on the front-end.
- I've allowed a content model's association with a taxonomy to be removed _even if it's the last model associated to that taxonomy_. This bypasses the front-end requirement that forces users to choose at least one model for every taxonomy, but I didn't see any good alternatives (i.e. preventing the model deletion or deleting the whole taxonomy if this was the last model). I wonder if dropping the front-end requirement might be the way to go?
     - **Another consequence of this decision:** After deleting all of your models, taxonomies might still exist in the DB but remain inaccessible from the UI due to the fact that we only show the "View Taxonomies" button when you have at least one model. How we address this depends on what we do with the front-end requirement.

Related to the second point above, I started to feel like I was stretching the limits of our data structures here. What we have still works for now, but I wanted to go ahead and bring it up and start collecting thoughts in case others run into the same limitations. Feel free to chime in on [BH-1150](https://wpengine.atlassian.net/browse/BH-1150).

https://wpengine.atlassian.net/browse/BH-1148

## Testing

Unit and e2e tests are included. Manual steps:

1. Create Restaurant and Recipe models.
2. Create a Cuisines taxonomy and assign it to both Restaurants and Recipes.
3. Delete the Recipe model.
4. Click "View Taxonomies" to view the taxonomy list.
5. See that "restaurant" is the only model listed for Cuisines in the taxonomies table.

### DIFF
--- a/includes/rest-api/rest-api-endpoint-registration.php
+++ b/includes/rest-api/rest-api-endpoint-registration.php
@@ -13,6 +13,7 @@ use WP_Error;
 use WP_REST_Request;
 use function WPE\AtlasContentModeler\ContentRegistration\generate_custom_post_type_args;
 use function WPE\AtlasContentModeler\ContentRegistration\get_registered_content_types;
+use function WPE\AtlasContentModeler\ContentRegistration\Taxonomies\get_acm_taxonomies;
 use function WPE\AtlasContentModeler\ContentRegistration\update_registered_content_types;
 
 add_action( 'rest_api_init', __NAMESPACE__ . '\register_rest_routes' );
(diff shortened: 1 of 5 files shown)


matthewguywright (approved):

ChrisWiegman (approved):
LGTM

mindctrl (approved):

apmatthews:
Ticket for addressing orphaned taxonomies and other UI quirks that arise from this PR: [BH-1161](https://wpengine.atlassian.net/browse/BH-1161)

PR #209, opened 2021-07-28
[NO JIRA] Don't cache MySQL tables between requests

ChrisWiegman:
## Description

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -8,8 +8,7 @@ FROM wordpress:${WORDPRESS_VERSION}-php${PHP_VERSION}-apache
 RUN docker-php-ext-install pdo pdo_mysql
 
 # Add wp-cli to the main container.
-COPY --from=wp-cli --chown=www-data:www-data /usr/local/bin/wp /public/bin/wp-cli.phar
-COPY wp-cli/wp-cli.sh /usr/local/bin/wp
+COPY --from=wp-cli /usr/local/bin/wp /usr/local/bin/wp
 
 RUN set -e \
     ; apt-get -qq update \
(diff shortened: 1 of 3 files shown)


mindctrl:
@ChrisWiegman here's what happens on my local computer:

this branch:
`TEST=DeleteFieldFromModelCest:i_can_delete_a_field_from_a_model make test-e2e`
If I run this twice in a row, the second run fails because the model trying to be created already exists.

`main` branch:
`TEST=DeleteFieldFromModelCest:i_can_delete_a_field_from_a_model make test-e2e`
Same test passes every time I run it consecutively. 🤷‍♂️

nickcernis:
I see the same behaviour as John. Successive runs in this branch fail due to the already existing model (same command as John's example):

Screenshot 2021-08-03 at 08 52 09

Successive runs in `main` currently work for me:

```
…
Time: 14.14 seconds, Memory: 18.00 MB

OK (1 test, 3 assertions)
```

The combination of `dump` and `cleanup` in `acceptance.suite.yml` should already reset the database before each e2e test, in case that was the issue this was designed to address.

https://github.com/wpengine/atlas-content-modeler/blob/15a9a52548a193a16637040c70f300966d4ca072/tests/acceptance.suite.yml#L26-L28

> **dump** required - defaults to null; sets the path, relative to the project root folder, or absolute to the SQL dump file that will be used to set the tests initial database fixture. If set to null then the populate, cleanup and populator parameters will be ignored.
> **cleanup** - defaults to true empty the target database and import the SQL dump(s) specified in the dump argument before each test. 
> https://wpbrowser.wptestkit.dev/modules/wpdb

ChrisWiegman:
Closing as it is now stale

PR #210, opened 2021-07-29
BH-1095: Allow Developers to change option names

anthonyburchell:
## Description

This PR adds multiple choice fields back into ACM and adds better validation and a data structure that allows for content model creators to change choice names. The key value way we're storing them now should ensure that the slug is maintained and the name is dynamic.


includes/publisher/js/src/components/Field.jsx
```diff
@@ -196,19 +196,23 @@ function fieldMarkup(field, modelSlug, errors, validate) {
...
 										placeholder="Option Name"
+										value={item.name}
 										defaultChecked={
 											field.value &&
-											field.value.some(
-												(name) => name == item.name
-											)
+											field.value.some(function (slug) {
+												return (
+													Object.keys(slug) ==
```
matthewguywright, line 212: === vs ==

includes/settings/js/src/components/fields/MultipleChoiceFields.jsx
```diff
@@ -40,17 +54,17 @@ function MultipleChoiceFields({
 								return (
 									<div
 										key={item.id}
-										className="field multiple-option-container-single"
+										className={`choices[${index}].container field multiple-option-container-single`}
```
matthewguywright, line 57: That's an interesting class name. Do we want '.'s in the class name? Like would this be something like:
choices[1].container field ...

Or is this a WP standard thing?

includes/settings/js/src/components/fields/Form.jsx
```diff
@@ -210,6 +210,20 @@ function Form({ id, position, type, editing, storedData }) {
 						});
 					});
 				}
+				if (err.code === "wpe_option_slug_undefined") {
```
matthewguywright, line 213: It would be nice to see these errors broken out into a separate object or an enum and use a switch statement or external function instead. There are quite a few of these now.

includes/settings/js/src/components/fields/MultipleChoiceFields.jsx
anthonyburchell: it surprisingly works, but yeah I think we could switch these to `-` between each word. `choices-1-container`

nickcernis: If we're using that class only as a selector for the “remove choice” e2e test, another option is to:

1. Remove the className here:

```suggestion
										className={`field multiple-option-container-single`}
```
2. Change the test selector:

```diff
diff --git a/tests/acceptance/CreateContentModelMultipleChoiceFieldCest.php b/tests/acceptance/CreateContentModelMultipleChoiceFieldCest.php
index 3fc3613..26fe3f3 100644
--- a/tests/acceptance/CreateContentModelMultipleChoiceFieldCest.php
+++ b/tests/acceptance/CreateContentModelMultipleChoiceFieldCest.php
@@ -28,7 +28,7 @@ class CreateContentModelMultipleChoiceFieldCest
         $I->wait(1);
         $I->seeInField('choices[0].name', 'dog');
         $I->see('Editing “Favorite Animal” Field', '.field-list');
-        $I->click('.remove-option', '.choices[0].container');
+        $I->click('Remove choice');
         $I->wait(1);
         $I->seeInField('choices[0].name', 'cat');
     }
```

The test runner will click the first button with the “Remove choice” text, which is the same button that was previously identified by `choices[0].container`.

includes/settings/js/src/components/fields/MultipleChoiceFields.jsx
```diff
@@ -7,6 +7,7 @@ import supportedFields from "./supportedFields";
 import AddIcon from "../../../../../components/icons/AddIcon";
 import TrashIcon from "../../../../../components/icons/TrashIcon";
 import Icon from "../../../../../components/icons";
+import { toValidApiId } from "./toValidApiId";
```
nickcernis, line 10: Just a heads up that Andrew's PR at https://github.com/wpengine/atlas-content-modeler/pull/214 changes the location of `toValidApiId`. (Just something to look out for in case that PR is merged first.)

nickcernis:
This is working well for me!

Apart from the above comments, I noticed that the `multipleChoice` field does not appear in WPGraphQL responses (I think because no type in `map_html_field_type_to_graphql_field_type` or resolver in `register_content_fields_with_graphql` are configured).

That might need to be a separate PR before we can ship this. (Unless we want to ship without WPGraphQL support and add it later?) REST support looks ok to me.

includes/publisher/lib/field-functions.php
```diff
@@ -102,7 +102,7 @@ function sanitize_field( string $type, $value ) {
 			if ( is_array( $value ) ) {
 				$options_object = [];
 				foreach ( $value as $option ) {
-					$options_object[] = key( $option );
+					$options_object[] = $option;
```
apmatthews, line 105: I may be totally missing something, but does this block actually change the value? The way it's written, I'm wondering if we need the `is_array()` conditional at all.

If we do want to transform this value somehow when it's an array, should we keep the `key()` function there as it was? Otherwise, we'll be saving option names. For instance, this is what would be saved for a multi-select field:

```
a:2:{i:0;a:1:{s:3:"two";s:3:"Two";}i:1;a:1:{s:5:"three";s:5:"Three";}}
```

(Unrelated to this PR, but it looks like we don't even use the value that gets returned from the `sanitize_field()` function 😬. I'll make a note to follow up on that!)

nickcernis: I agree that we could remove or adapt it. It was left when Repeater fields were removed:

- https://github.com/wpengine/atlas-content-modeler/pull/124#pullrequestreview-671211033
- https://github.com/wpengine/atlas-content-modeler/pull/124#issuecomment-850533862

mindctrl:
When adding this field to a model, the cursor focus goes to the name input for Choice 1, skipping the Field name input which appears first.

image

includes/publisher/js/src/components/Field.jsx
```diff
@@ -196,19 +196,23 @@ function fieldMarkup(field, modelSlug, errors, validate) {
...
 								>
 									{item.name}
 									<input
 										type="checkbox"
-										name={`atlas-content-modeler[${modelSlug}][${field.slug}][${index}][${item.name}]`}
-										id={`atlas-content-modeler[${modelSlug}][${field.slug}][${index}][${item.name}]`}
+										name={`atlas-content-modeler[${modelSlug}][${field.slug}][${index}][${item.slug}]`}
+										id={`atlas-content-modeler[${modelSlug}][${field.slug}][${index}][${item.slug}]`}
 										placeholder="Option Name"
+										value={item.name}
```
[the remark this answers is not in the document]
nickcernis: Agreed, good catch.

nickcernis (commented):
@anthonyburchell I'm currently unable to create a multiple choice field. I see “Must create a choice first” when clicking Create, even after creating choices:

Screenshot 2021-10-25 at 10 57 49

Guessing that may be responsible to some of the e2e test failures on this branch?

I left another comment above about one of Andrew's suggestions, and dismissed some stale comments.

ChrisWiegman (commented):
We've chatted about this elsewhere but I feel it is important to bring it here before I approve this.

What about hiding the slug field for each choice? We're not using them in GraphQL and the requirement to create one manually is strange and confusing in the settings app. I feel we would be better serving users by autogenerating that in the background. How we do that, here or another ticket, is something I'm more than willing to discuss.

ChrisWiegman (dismissed):
Beyond the comments, I would like to see some tests for a GraphQL response (where appropriate).

mindctrl (changes requested):

mindctrl:
@anthonyburchell can you bring this up to date with `main`? Looks like quite a few things are missing, like recent Parcel fixes.

mindctrl:
Given a model like this screenshot:
image

---

And adding an entry like this screenshot:
image

---

Results in the meta value being stored like this:
`a:1:{i:0;a:1:{s:10:"greenBeans";s:10:"greenBeans";}}`

Does this seem right? Do we need to store the value twice for each selection? If I add another selection, the value looks like this:
`a:2:{i:0;a:1:{s:10:"greenBeans";s:10:"greenBeans";}i:1;a:1:{s:4:"rice";s:4:"rice";}}`

matthewguywright:
General Notes:

1. Remove choice is awkward, especially in smaller screen resolutions for tablets and smaller. Is the text necessary? Can it use just the red trash can or a red "x"?
2. Currently it doesn't auto generate the identifier
3. Character count for "Name" field is misaligned with the input
4. Label not aligned vertically in publisher with the radio and checkbox input
5. Legend label looks indented by 2px from the rest of the form

includes/rest-api/routes/content-model-field.php
```diff
@@ -126,14 +125,69 @@ function dispatch_update_content_model_field( WP_REST_Request $request ) {
...
+		$problem_option_slug_index = [];
+		foreach ( $params['choices'] as $index => $choice ) {
+			if ( content_model_multi_option_slug_exists( $params['choices'], $choice['slug'], $index ) ) {
+				$problem_option_slug_index[] = $index;
+			}
+		}
+		if ( $problem_option_slug_index ) {
+			$problem_duplicate_slug = new WP_Error(
+				'wpe_option_slug_duplicate',
+				'Another choice in this field has the same API identifier.',
```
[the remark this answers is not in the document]
nickcernis: Confirmed fixed in 44440f7c30a81f1d4045b153a22a8075f2e4bb2a.

includes/rest-api/routes/content-model-field.php
```diff
@@ -126,14 +125,69 @@ function dispatch_update_content_model_field( WP_REST_Request $request ) {
...
+		$problem_name_index = [];
+		foreach ( $params['choices'] as $index => $choice ) {
+			if ( content_model_multi_option_exists( $params['choices'], $choice['name'], $index ) ) {
+				$problem_name_index[] = $index;
+			}
+		}
+		if ( $problem_name_index ) {
+			$problem_duplicate_name = new WP_Error(
+				'wpe_duplicate_content_model_multi_option_id',
+				'Another choice in this field has the same name.',
```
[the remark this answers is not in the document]
nickcernis: Confirmed fixed in 44440f7c30a81f1d4045b153a22a8075f2e4bb2a.

includes/rest-api/routes/content-model-field.php
```diff
@@ -126,14 +125,69 @@ function dispatch_update_content_model_field( WP_REST_Request $request ) {
...
+		$problem_index = [];
+		foreach ( $params['choices'] as $index => $choice ) {
+			if ( $choice['slug'] === '' ) {
+				$problem_index[] = $index;
+			}
+		}
+		if ( ! empty( $problem_index ) ) {
+			$problem_error_slug_blank = new WP_Error(
+				'wpe_option_slug_undefined',
+				'Multiple Choice Field update failed, please set a slug for your choice before saving.',
```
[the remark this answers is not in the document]
nickcernis: Confirmed fixed in 44440f7c30a81f1d4045b153a22a8075f2e4bb2a.

mindctrl (dismissed):

includes/settings/js/src/components/fields/Form.jsx
mindctrl: Agreed, but I think we can tackle that in the future.

includes/rest-api/routes/content-model-field.php
```diff
@@ -104,43 +105,96 @@ function dispatch_update_content_model_field( WP_REST_Request $request ) {
 	if ( isset( $params['type'] ) && $params['type'] === 'multipleChoice' && ! $params['choices'] ) {
 		return new WP_Error(
 			'wpe_invalid_multi_options',
-			'Multiple Choice update failed. Options need to be created before updating a Multiple Choice field.',
+			__( 'Multiple Choice update failed. Options need to be created before updating a Multiple Choice field.', 'atlas-content-modeler' ),
```
mindctrl, line 108:
```suggestion
			__( 'Multiple Choice update failed. Choices need to be created before updating a Multiple Choice field.', 'atlas-content-modeler' ),
```

Should we say "Choices" instead of "Options" here, since the UI calls them choices?

includes/settings/js/src/components/fields/MultipleChoiceFields.jsx
```diff
@@ -167,8 +270,10 @@ function MultipleChoiceFields({
 												<span className="error">
 													<Icon type="error" />
 													<span role="alert">
-														Cannot have duplicate
-														option names.
+														{__(
+															"Cannot have duplicate choice names.",
+															"atlas-content-modeler"
```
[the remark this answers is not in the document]
nickcernis: I agree with this, but perhaps to keep the PR on track we could handle updating error message handling in a separate PR? There are some other places we should change it, as @anthonyburchell noted in standup.

includes/rest-api/routes/content-model-field.php
```diff
@@ -104,43 +105,96 @@ function dispatch_update_content_model_field( WP_REST_Request $request ) {
...
+				$problem_name_index[] = $index;
+			}
+		}
+		if ( $problem_name_index ) {
+			$problem_duplicate_name = new WP_Error(
+				'wpe_duplicate_content_model_multi_option_id',
+				__( 'Another choice in this field has the same name.', 'atlas-content-modeler' ),
+				array( 'status' => 400 )
+			);
+			$problem_duplicate_name->add( 'problem_name_index', $problem_name_index );
```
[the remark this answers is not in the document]
nickcernis: Addressed in 866bbfb35be5dcc76f7d21d029279fc8d43f9935 and 2543ab9dc6c5a98e8aa98dcba9c86c513a3a4518.

nickcernis (dismissed):
Looks good! 🥳 

Noting that we chose not to auto-generate the choice slugs from their names for now. (This requires a refactor of the `useInputGenerator` hook to support dynamically sized arrays of inputs, which seems out of scope for now.)

nickcernis (approved):

mindctrl (approved):

PR #213, opened 2021-08-02
BH-1143: Fix negative numbers min vs max

matthewguywright:
## Description

Update number type to not allow entering numbers less than min or more than max including handling negative numbers.

https://wpengine.atlassian.net/browse/BH-1143

## Testing

1. Check for number required
2. Check for number not required
3. Check for input less than min
4. Check for input more than max


includes/settings/js/src/components/fields/AdvancedSettings.jsx
```diff
@@ -1,7 +1,7 @@
 /**
  * Additional field options that appear in the Advanced Settings modal.
  */
-import React from "react";
+import React, { useRef } from "react";
```
nickcernis, line 4: `useRef` is introduced here but not used.

```suggestion
import React from "react";
```

tests/acceptance/PublishModelCest.php
```diff
@@ -79,6 +79,98 @@ public function i_see_no_submission_errors_in_number_fields_when_input_is_missin
...
+		$i->wait(1);
+
+		$i->fillField(['name' => 'atlas-content-modeler[goose][integer]'], '-1');
+		$i->fillField(['name' => 'atlas-content-modeler[goose][decimal]'], '-1');
+		$i->scrollTo('#submitdiv');
+
+		$i->click('Publish', '#publishing-action');
+		$i->wait(2);
+
+		$i->see('Minimum value is');
```
nickcernis, line 121: Might be worth checking both fields for the error:

```suggestion
		$i->see('Minimum value is', '#field-integer');
		$i->see('Minimum value is', '#field-decimal');
```

Otherwise one field could fail to display the error but the test would still pass.

tests/acceptance/PublishModelCest.php
```diff
@@ -79,6 +79,98 @@ public function i_see_no_submission_errors_in_number_fields_when_input_is_missin
...
+		$i->wait(1);
+
+		$i->fillField(['name' => 'atlas-content-modeler[goose][integer]'], '20');
+		$i->fillField(['name' => 'atlas-content-modeler[goose][decimal]'], '20');
+		$i->scrollTo('#submitdiv');
+
+		$i->click('Publish', '#publishing-action');
+		$i->wait(2);
+
+		$i->see('Maximum value is');
```
nickcernis, line 167: Same recommendation here:

```suggestion
		$i->see('Maximum value is', '#field-integer');
		$i->see('Maximum value is', '#field-decimal');
```

nickcernis (approved):

mindctrl (approved):

PR #214, opened 2021-08-02
[BH-1147] Format Model IDs as WordPress post type slugs

apmatthews:
## Description

The primary goal of this PR is to standardize how we format model slugs (also known as Model IDs). All model slugs eventually become WordPress post type slugs. Using the same formatting rules for model slugs that WP uses for post types should eliminate or reduce the need to reformat model slugs before using them to reference post types.

#### The REST routes for creating/updating content models now format model slugs with `sanitize_key()` before saving.

This is primarily handled through updates to the `create_model()` function. It's also worth noting that the `create_model()` function now returns the created model rather than a boolean so that clients know if `sanitize_key()` has resulted in a different model slug than what was submitted.

Model update routes did not require any special consideration because changing a model slug is currently unsupported. However, there is a small, one line change that makes sure this limitation is enforced.

#### The DevEx application formats model slugs to match the format defined by the WordPress `sanitize_key()` function.

This was accomplished by refactoring the `useApiIdGenerator` hook to be more generic. The hook is now named `useInputGenerator` and allows the client to supply their own formatting function. As a result, we can use `toValidApiId()` to generate field slugs and `toPostTypeSlug()` to generate model slugs.

#### Any references to model slugs have been updated to account for the new format.

- Several cases of lowercasing a model slug before it is used have been removed.
- Model and Field REST routes that include a model slug have an updated regex that matches `sanitize_key()` style slugs.

#### Backwards Compatibility

It's fully backwards compatible! We were consistently using a wrapper function around the `get_option()` call for retrieving models, so that was a natural place to reformat existing model slugs. ~~Since all of our models are stored together in a single option, anything that triggers an option write in the plugin (i.e. creating new models, updating or adding fields to models) will automatically correct all of the unformatted slugs in the DB.~~ We are now persisting updated model slugs on retrieval if any of the existing slugs are changed. I've included a TODO note to consider removing this compatibility patch when we move to v1.0 so that we're not stuck with it forever.

I also explored the idea of including a very simple migration system, but scrapped it in favor of the lazy update approach. We'll likely want to revisit the idea at some point, but such a task felt like it deserved its own separate PR and suite of tests.

https://wpengine.atlassian.net/browse/BH-1147

## Testing

- PHP unit tests are included.
- Unit test for the `toPostTypeSlug()` formatter included.
- I had trouble getting Jest tests written for the `useInputGenerator` hook, but I think a majority of that functionality is covered in existing e2e tests.

#### Manual testing steps focused on backwards compatibility

1. Before checking out this branch, create the following three models entering only the singular/plural labels and taking the defaults for everything else:
    1. Rabbit/Rabbits
    2. Two Word/Two Words
    3. Special-Character_9/Special-Characters_9
2. Check out this branch and rebuild the plugin.
3. Refresh the WordPress dashboard and see that all three models you created still exist.
4. Look at the `atlas_content_modeler_post_types` option in your database and verify that all models slugs have been lowercased (or more specifically, contain only lowercase alphanumeric characters, dashes, and underscores)
5. Verify that model creation, model editing, and model field creation still works.

## Dependant PRs

The `taxonomies` branch will need to be updated once this has been merged. I will take care of that reconciliation.


includes/rest-api/rest-api-endpoint-registration.php
```diff
@@ -166,7 +166,6 @@ function dispatch_create_content_model( WP_REST_Request $request ) {
 
 	unset( $params['_locale'] ); // Sent by wp.apiFetch but not needed.
 
-	// @todo simplify create_model() signature to single array?
```
apmatthews, line 169: Removed this todo because I think keeping the model slug as an explicit argument is valuable. The slug is a model's unique identifier, so we want to be pretty strict about requiring it. I don't think you should be able to call the `create_model()` function without it.

tests/integration/content-registration/test-rest-content-model-endpoint.php
```diff
@@ -77,20 +100,34 @@ public function test_cannot_update_model_with_invalid_data(): void {
 		wp_set_current_user( 1 );
 		$model   = 'rabbits';
 
-		// First request to create model.
```
apmatthews, line 80: Several of the tests in this file were firing a REST request to create a model before testing update requests. This no longer seemed necessary. In every case, the model had already been created in the test's `_before()` step.

includes/rest-api/rest-api-endpoint-registration.php
```diff
@@ -555,7 +556,7 @@ function create_model( string $post_type_slug, array $args ) {
 		return new WP_Error( 'model-not-created', esc_html__( 'Model not created. Reason unknown.', 'atlas-content-modeler' ) );
 	}
 
-	return true;
+	return $content_types[ $post_type_slug ];
```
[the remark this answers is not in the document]
apmatthews: Can't deny that there are quite a few returns in this function, but this PR doesn't add any. Anything we can do to ignore the suggestion in this case? @ChrisWiegman

includes/content-registration/custom-post-types-registration.php
```diff
@@ -311,7 +311,30 @@ function generate_custom_post_type_args( array $args ): array {
...
  */
 function get_registered_content_types(): array {
-	return get_option( 'atlas_content_modeler_post_types', array() );
+	$models = get_option( 'atlas_content_modeler_post_types', array() );
+
+	/**
+	 * Maintains backwards compatibility with models that were created
+	 * before sanitize_key() was used to format model slugs on creation.
+	 *
+	 * Exisitng data will be lazily corrected as models are added or
```
nickcernis, line 320: Tiny typo:

```suggestion
	 * Existing data will be lazily corrected as models are added or
```

includes/rest-api/rest-api-endpoint-registration.php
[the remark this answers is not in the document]
nickcernis: Happy to ignore this. I don't think multiple returns are an issue when used as [guard clauses](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html). (The `create_model` function itself might be ready to refactor/shorten, but not as a result of this PR.)

includes/rest-api/rest-api-endpoint-registration.php
```diff
@@ -591,6 +592,9 @@ function update_model( string $post_type_slug, array $args ) {
 
 	$new_args = wp_parse_args( $args, $content_types[ $post_type_slug ] );
 
+	// Updating the slug is unsupported.
+	$new_args['slug'] = $post_type_slug;
```
nickcernis, line 596: Nice touch!

includes/settings/js/src/hooks/useInputGenerator.js
```diff
@@ -0,0 +1,56 @@
+import { useState } from "react";
+
+/**
+ * Uses the value of a source input to generate and apply a valid value
+ * to a another input.
```
nickcernis, line 5:
```suggestion
 * to another input.
```

nickcernis (dismissed):
Working beautifully!

- I like the approach of renaming/refactoring `useInputGenerator` to accept different format functions. Thanks for extending the test cases to the new formatter.
- I confirmed model slugs become sanitized when a model is edited without impact to pre-existing entries for that post type (`twoWord => twoword` and `specialCharacter_9` => `specialcharacter_9`). I like that this potentially removes the need for a migration. (I suspect we would need to introduce a migration if we ever want to remove the “maintains backwards compatibility” logic introduced here, though. We've removed the `strtolower` and `STR_LOWER` calls/filters here and it probably isn't safe to assume that everyone will have made an edit to every affected model to lowercase existing model IDs. Something we can think about pre-1.0 before we make any decisions to remove the backcompat code, or defer until we do have a migration process in place.)
- Noting that the relocation of `toValidApiId` will affect @anthonyburchell's PR which also [makes use of that function](https://github.com/wpengine/atlas-content-modeler/pull/210/files#diff-05b60f91245688634835742ddc86df1cfa676633a70df32a7637f2918a9b0b9dR10). (We'll just need to update the path here or there depending on which PR is merged first.)

includes/content-registration/custom-post-types-registration.php
```diff
@@ -311,7 +311,30 @@ function generate_custom_post_type_args( array $args ): array {
...
+	 *
+	 * @todo Consider removing before v1.0.
+	 */
+	$updated_models = [];
+	foreach ( $models as $key => $model ) {
+		$slug = sanitize_key( $key );
+
+		if ( $key !== $slug ) {
+			$model['slug']   = $slug;
+			$models[ $slug ] = $model;
```
mindctrl, line 331: Do we need to update `$models` here? It's not referenced outside of this loop.

includes/content-registration/custom-post-types-registration.php
mindctrl: Technically this data is corrected at runtime when the data is fetched, it just isn't persisted until add/edit. It seems yucky, but I wonder if we should do the update here... 🤔  That would make it easier to remove later, because we'd know that if someone had models and the site had even 1 hit where the `init` hook fired, the data would be updated... which would cover every site. This would work around the concern @nickcernis had with removing this code in v1.0 if someone hadn't edited an existing model. Thoughts?

apmatthews: Just to be completely clear, ALL models are corrected whenever ANY model is added/edited. It's not only the model you are editing that is fixed. Each of the following actions would persist a fully updated set of data:

1. Creating/Updating/Deleting a new model.
2. Updating/Deleting an existing model.
3. Creating/Updating/Deleting fields on an existing model.
4. Creating/Updating/Deleting fields on a new model.

That means the only case where updated data is not persisted is if you are no longer making changes of any kind to your content models. Granted, this very well could be the case for some users.

I'm coming around to the idea of persisting the updates during retrieval as my gag reflex begins to subside 😆. Like you said, doing so would make the case for removing the patch pre v1.0 much stronger, which I see as a plus. I can wrap the update in a conditional so that it only runs once on the sites that need it.

includes/content-registration/custom-post-types-registration.php
apmatthews: Nope! Good catch. That's left over from some refactoring. I'm revisiting this block to persist updates anyway, so I'll correct it there.

apmatthews:
@nickcernis @mindctrl This is ready for another look. I've updated the backwards compatibility patch to include writing the updated data back to the database during retrieval. That should make it easier to remove the patch in the future.

PR description and manual testing steps have been updated accordingly.

mindctrl (approved):

nickcernis (approved):
Looking good. I confirmed the updated `get_registered_content_types` logic is working to alter model slugs on fetch.

> I've updated the backwards compatibility patch to include writing the updated data back to the database during retrieval. That should make it easier to remove the patch in the future.

Agreed. The remaining minor risks in removal are:
- Sites that haven't run `get_registered_content_types` between this update and our 1.0. Low risk, though, as we call that function on `init` in `register_content_types()`.
- Sites that jump versions from 0.4.1 (current version without the slug migration logic) to 1.0 (or whenever the migration code is removed). Also low risk, but just noting the possibility in case it comes up in support.

PR #216, opened 2021-08-03
[BH 1144] Refactor testing

ChrisWiegman:
## Description

This PR refactors API testing to group the tests in a more logical fashion and to utilize a complete and consistent collection of model and field data throughout. The idea here is that can validate all endpoints for both GraphQL and REST within PhpUnit. These are much faster than the other suits which we will still use to supplement overall functionality (can we get to the endpoints, etc) without validating the data (we can determine data constraints, permissions and other items here).

Note: this largely moves existing tests to use a shared set of complete models and fields. Each model and field represents what can be generated and available in the WordPress database with a complete parameter set. Assertions were adjusted, where needed, to meet the new data set but, as of now, no new tests have been added. This PR will require a few follow-ups:

* Remove duplicate tests from Jest and CodeCeption
* Implement high-level testing in Jest to confirm the overall functionality of both the REST and GraphQL APIs. 
* Add further tests and assertions to better validate field manipulation as well as constraints on any CRUD functions performed within the API.

For the most part our existing tests were complete, if not a little scattered. There were a few bugs found for which I've opened the appropriate JIRA tickets. In addition, we test heavily for `show_in_graphql` and its rest counterpart however neither field seems to be used or controllable throughout. Setting a private field does not relect these params. As a result I've updated existing tests to correctly reflect both public and private fields we can create but it would be worth another pass to better handle or eliminate checks of these particular attributes.


nickcernis (approved):
Thanks for this, Chris. I like the thoroughness, and that it makes it easier to maintain a suite of tests as we add new field types.

mindctrl (approved):

PR #218, opened 2021-08-04
[NO JIRA] Apply slug sanitization updates to taxonomies

apmatthews:
## Description

Accounting for the merge of #214:

- Replaces the `useApiIDGenerator` hook with `useInputGenerator`.
- Sanitizes taxonomy slugs server-side.
- Enforces slug length restrictions server-side.
- Changes API Identifier label to Taxonomy ID (following pattern from #200)
- Changes formatting of Taxonomy ID to match `sanitize_key()` behavior.
    - WordPress doesn't do this sanitization automatically on registration of taxonomies like they do for post types, so we wouldn't technically run into any mismatch between stored and registered taxonomy slug values. However, sanitizing still seems like a good thing for us to do. The taxonomy slug is used as an array key, so sanitize_key() still seemed like a good choice.

## Testing

Existing tests should provide sufficient coverage for these changes.

### DIFF
--- a/includes/rest-api/rest-api-endpoint-registration.php
+++ b/includes/rest-api/rest-api-endpoint-registration.php
@@ -797,7 +797,7 @@ function save_taxonomy( array $params, bool $is_update ) {
 	if ( in_array( $params['slug'], $non_acm_taxonomies, true ) ) {
 		return new WP_Error(
 			'atlas_content_modeler_taxonomy_exists',
-			esc_html__( 'A taxonomy with this API Identifier already exists.', 'atlas-content-modeler' ),
+			esc_html__( 'A taxonomy with this Taxonomy ID already exists.', 'atlas-content-modeler' ),
 			[ 'status' => 400 ]
 		);
 	}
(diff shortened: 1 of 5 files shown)


includes/rest-api/rest-api-endpoint-registration.php
```diff
@@ -781,10 +781,13 @@ function content_model_multi_option_exists( array $names, string $current_choice
  * @since 0.6.0
  */
 function save_taxonomy( array $params, bool $is_update ) {
-	if ( empty( $params['slug'] ) ) {
+	// Sanitize key allows hyphens, but it's close enough to register_taxonomy() requirements.
+	$params['slug'] = isset( $params['slug'] ) ? sanitize_key( $params['slug'] ) : '';
+
+	if ( empty( $params['slug'] ) || strlen( $params['slug'] ) > 32 ) {
```
mindctrl, line 787: ❤️ Love the strlen check

mindctrl (approved):

apmatthews:
@mindctrl Just added 285c83c. Not a critical change, but is something I had planned to circle back to. Apologies for the re-ping!

nickcernis (approved):
Thanks for taking care of this!

includes/settings/js/src/components/Taxonomies.jsx
```diff
@@ -26,7 +26,7 @@ export default function Taxonomies() {
 		}
 
 		const getTaxonomyEditHref = (slug, type) =>
-			`edit-tags.php?taxonomy=${slug}&post_type=${type.toLowerCase()}`;
+			`edit-tags.php?taxonomy=${slug}&post_type=${type}`;
```
nickcernis, line 29: Great addition, nice to be able to get rid of these.

PR #222, opened 2021-08-11
[BH-1135] Add Publisher UI for 1:1 Relationships

ChrisWiegman:
## Description


anthonyburchell:
One note about this PR, the table styling is off due to some color vars that are currently in Main and not this branch. I could rebase, or we can clean that up as part of the one to many relationships.

includes/publisher/js/src/components/Field.jsx
```diff
@@ -77,6 +80,149 @@ export default function Field(props) {
 
 function fieldMarkup(field, modelSlug, errors, validate) {
 	switch (field.type) {
+		case "relationship":
+			const [
```
nickcernis, line 84: Should we consider breaking this out into its own component to make the `switch` more readable? 

Not blocking because we've done this inconsistently so far (only for the `media` field), so it could be part of future refactoring work.

includes/publisher/js/src/components/Field.jsx
```diff
@@ -77,6 +80,149 @@ export default function Field(props) {
...
+								relationshipClickHandler(
+									e,
+									field,
+									modelSlug,
+									errors,
+									validate
+								);
+							}}
+						>
+							{/*<Icon type="error" />*/}
```
nickcernis, line 210: Based on the mockup it looks like we need to add a “link” icon here, unless this is commented out now on purpose for future styling work.

includes/publisher/js/src/components/Field.jsx
```diff
@@ -77,6 +80,149 @@ export default function Field(props) {
...
+					}
+
+					return response.json();
+				});
+			}
+
+			/**
+			 * Gets entries whenever the state of 'page' changes.
+			 * Caches those entries in the pagedEntries object, keyed by page.
+			 */
```
nickcernis, line 130: Comment text seems wrong (looks copied from the useEffect from the modal?).

includes/publisher/js/src/components/Field.jsx
```diff
@@ -77,6 +80,149 @@ export default function Field(props) {
...
+							id={`atlas-content-modeler[${modelSlug}][${field.slug}]`}
+							onClick={(e) => {
+								e.preventDefault();
+								relationshipClickHandler(
+									e,
+									field,
+									modelSlug,
+									errors,
+									validate
+								);
```
nickcernis, line 207: Could this just be `setEditSingleRelModalIsOpen(true);`? Then we could remove the `relationshipClickHandler` function.

We don't seem to use `relationshipClickHandler` elsewhere. It doesn't use the arguments we're passing here, and it's a one-liner so probably a good candidate to inline.

includes/publisher/js/src/components/Field.jsx
```diff
@@ -77,6 +80,149 @@ export default function Field(props) {
...
+					{field.value &&
+						relatedContent?.map((entry) => {
+							const { id, title } = entry;
+							return (
+								<>
+									<div className="app-card">
+										<section className="card-content">
+											<ul className="model-list">
+												<li>
+													{console.log(field.value)}
```
nickcernis, line 165: console.log

includes/publisher/js/src/components/RelationshipModal.jsx
```diff
@@ -0,0 +1,297 @@
...
+			)}
+			<div className="d-flex flex-row-reverse mt-2">
+				<button
+					type="submit"
+					disabled={typeof selectedEntry === "undefined"}
+					className="action-button mx-3"
+					onClick={(event) => {
+						event.preventDefault();
+						// TODO: Update the reference field's value here.
+						console.log(`Saving field ${selectedEntry}.`);
```
nickcernis, line 276: Are these still needed?

includes/publisher/js/src/components/RelationshipModal.jsx
```diff
@@ -0,0 +1,297 @@
...
+ * @param {Function} setIsOpen - Callback for opening and closing modal.
+ * @returns {JSX.Element} Modal
+ */
+export default function RelationshipModal({ field, isOpen, setIsOpen }) {
+	const [page, setPage] = useState(1);
+	const [pagedEntries, setPagedEntries] = useState({});
+	const [totalEntries, setTotalEntries] = useState(0);
+	const [selectedEntry, setSelectedEntry] = useState(); // TODO: set initial state value from stored field value.
+	const [selectedEntryTitle, setSelectedEntryTitle] = useState(); // TODO: set initial state value from stored field value.
+	const entriesPerPage = 5; // TODO: change to 5. 1 is for testing pagination only.
```
nickcernis, line 24: TODO looks done.

includes/publisher/js/src/components/RelationshipModal.jsx
```diff
@@ -0,0 +1,297 @@
...
+ * @param {Boolean} isOpen Whether or not the model is open.
+ * @param {Function} setIsOpen - Callback for opening and closing modal.
+ * @returns {JSX.Element} Modal
+ */
+export default function RelationshipModal({ field, isOpen, setIsOpen }) {
+	const [page, setPage] = useState(1);
+	const [pagedEntries, setPagedEntries] = useState({});
+	const [totalEntries, setTotalEntries] = useState(0);
+	const [selectedEntry, setSelectedEntry] = useState(); // TODO: set initial state value from stored field value.
+	const [selectedEntryTitle, setSelectedEntryTitle] = useState(); // TODO: set initial state value from stored field value.
```
nickcernis, line 23: `setSelectedEntryTitle` is called but `selectedEntryTitle` is never used. It looks like this could be removed.

includes/publisher/js/src/components/RelationshipModal.jsx
```diff
@@ -0,0 +1,297 @@
...
+ * @param {Object} field The relationship field.
+ * @param {Boolean} isOpen Whether or not the model is open.
+ * @param {Function} setIsOpen - Callback for opening and closing modal.
+ * @returns {JSX.Element} Modal
+ */
+export default function RelationshipModal({ field, isOpen, setIsOpen }) {
+	const [page, setPage] = useState(1);
+	const [pagedEntries, setPagedEntries] = useState({});
+	const [totalEntries, setTotalEntries] = useState(0);
+	const [selectedEntry, setSelectedEntry] = useState(); // TODO: set initial state value from stored field value.
```
nickcernis, line 22:
```suggestion
	const [selectedEntry, setSelectedEntry] = useState(parseInt(field?.value));
```

This ensures that the currently selected entry will have its radio button active when you load an existing post and click “link new reference”.

includes/publisher/js/src/components/RelationshipModal.jsx
```diff
@@ -0,0 +1,297 @@
...
+													modified
+												)}
+											</label>
+										</td>
+									</tr>
+								);
+							})}
+						</tbody>
+					</table>
+					<div className="d-flex flex-row-reverse">
```
nickcernis, line 185: Should we only show pagination info and buttons if `totalPages > 1`? (I think there may be a case to at least hide the buttons.)

includes/publisher/js/src/components/Field.jsx
```diff
@@ -77,6 +80,149 @@ export default function Field(props) {
...
+					<label
+						htmlFor={`atlas-content-modeler[${modelSlug}][${field.slug}]`}
+					>
+						{field.name}
+					</label>
+					{field.value &&
+						relatedContent?.map((entry) => {
+							const { id, title } = entry;
+							return (
+								<>
```
nickcernis, line 160: The `<>` [fragment](https://reactjs.org/docs/fragments.html) is unneeded here. (The div below it is a single element — the fragment is only needed to return multiple divs as one that otherwise would not want an extra <div> to wrap them.)

The div also needs a `key` to prevent “Each child in a list should have a unique "key" prop.” that's appearing in the console at the moment. (For example: `<div key={id} className="app-card">`.)

includes/publisher/js/src/components/RelationshipModal.jsx
```diff
@@ -0,0 +1,297 @@
...
+			<div className="d-flex flex-row-reverse mt-2">
+				<button
+					type="submit"
+					disabled={typeof selectedEntry === "undefined"}
+					className="action-button mx-3"
+					onClick={(event) => {
+						event.preventDefault();
+						// TODO: Update the reference field's value here.
+						console.log(`Saving field ${selectedEntry}.`);
+						field.value = `${selectedEntry}`;
```
nickcernis, line 277: Even though this might work, it's considered an anti-pattern in React to modify passed props directly (you no longer benefit from React features such as batched updates):

https://reactjs.org/docs/components-and-props.html#props-are-read-only

If we need to change state that was passed in from a parent component, we can:

- Create or reuse a `useState` in Fields.jsx.
- In Fields.jsx, pass the callback function from that useState to the RelationshipModal as a prop.
- Use that passed callback (`setActiveField()` or `setRelatedContent()`) or whatever it ends up being in RelationshipModal in place of `field.value = …` here.

includes/publisher/scss/_relationship-modal.scss
```diff
@@ -0,0 +1,6 @@
+@import "../../../includes/shared-assets/scss/_variables";
+
+h2 {
```
nickcernis, line 3: Needs to be scoped to the modal. It's currently breaking the WP publish box:

Screenshot 2021-08-18 at 14 08 42

includes/publisher/js/src/components/Field.jsx
```diff
@@ -77,6 +80,149 @@ export default function Field(props) {
...
+									<div className="app-card">
+										<section className="card-content">
+											<ul className="model-list">
+												<li>
+													{console.log(field.value)}
+													<div className="relation-model-card flex-wrap d-flex flex-column d-sm-flex flex-sm-row">
+														<span className="flex-item mb-3 mb-sm-0 pr-1">
+															<p className="label">
+																Linked Reference
+															</p>
```
nickcernis, line 170: Do we need this title? (Only asking because it's not in the mockup.)

| Current version | Mockup |
|-|-|
| Screenshot 2021-08-18 at 14 19 00 | Screenshot 2021-08-18 at 14 18 51 |

If we think we need it, it needs to be translatable.

includes/publisher/js/src/components/Field.jsx
```diff
@@ -77,6 +80,150 @@ export default function Field(props) {
...
+			 * Gets post information to display to the user outside of the modal.
+			 */
+			useEffect(() => {
+				if (field.value !== "") {
+					// let's not make the call if we don't have to.
+					getRelatedTitles(field).then((entries) => {
+						setRelatedContent(() => {
+							return entries;
+						});
+					});
```
nickcernis: If you want to, you can simplify this to:

```js
					getRelatedTitles(field).then((entries) => {
						setRelatedContent(entries);
					});
```

(Passing a function to a `useState` callback like `setRelatedContent` is useful when we want to extend or alter the existing state. If we're just replacing state it's fine to pass the new state instead of a function.)

That can be simplified further to the line below (`entries` is passed to `setRelatedContent` so there's no need for an anonymous function — we can just pass the function reference). No obligation to use it but feel free to if you think it's more readable:

```js
					getRelatedTitles(field).then(setRelatedContent);
```

(Not trying to code golf this, by the way — just reducing code if it can help legibility.)

Whenever you see:

```js
(a) => { myFunctionName(a); }
```

It's normally a sign it can be simplified to:

```js
myFunctionName
```

(Note lack of `()` — we are passing a reference to the function and not calling the function with `myFunctionName()`.)

nickcernis: Refactored in https://github.com/wpengine/atlas-content-modeler/pull/222/commits/8cfd503c611097d1aae495621401185ef52560c1 and https://github.com/wpengine/atlas-content-modeler/pull/222/commits/eafd4cc1afc8f8b431fd928bc1d5490746af1d42.

includes/publisher/js/src/components/RelationshipModal.jsx
```diff
@@ -0,0 +1,316 @@
...
+ * @param {Object} field The relationship field.
+ * @param {Boolean} isOpen Whether or not the model is open.
+ * @param {Function} setIsOpen - Callback for opening and closing modal.
+ * @returns {JSX.Element} Modal
+ */
+export default function RelationshipModal({ field, isOpen, setIsOpen }) {
+	const [page, setPage] = useState(1);
+	const [pagedEntries, setPagedEntries] = useState({});
+	const [totalEntries, setTotalEntries] = useState(0);
+	const [selectedEntry, setSelectedEntry] = useState(field.value.split(",")); // TODO: set initial state value from stored field value.
```
nickcernis, line 22:
```suggestion
	const [selectedEntry, setSelectedEntry] = useState(field.value.split(","));
```

includes/publisher/js/src/components/Field.jsx
ChrisWiegman: Good catch, Nick. I've removed the extra title

includes/publisher/js/src/components/RelationshipModal.jsx
nickcernis: Also, should it be `selectedEntries` now to account for one-to-many?

includes/publisher/js/src/components/Field.jsx
ChrisWiegman: Don't we have a ticket to break these out more in general? If so this and others should be a subtask of that

includes/publisher/js/src/components/Field.jsx
nickcernis: Nice! I think that looks cleaner, particularly for the one-to-many field, where “Linked Reference” would have been repeated.

Screenshot 2021-08-18 at 15 59 54

includes/publisher/scss/_buttons.scss
```diff
@@ -0,0 +1,167 @@
+@import "../../../includes/shared-assets/scss/_variables";
+button {
```
nickcernis, line 2: A heads-up that these also need to be scoped/prefixed. They're affecting WP-native modal styling, such as in the Media popover:

Screenshot 2021-08-18 at 16 13 53

includes/publisher/js/src/components/RelationshipModal.jsx
```diff
@@ -0,0 +1,316 @@
+/* global atlasContentModelerFormEditingExperience */
+import React, { useContext, useEffect, useState } from "react";
+import { sprintf, __ } from "@wordpress/i18n";
+import Modal from "react-modal";
+const { wp } = window;
+const { date, apiFetch } = wp;
+
+Modal.setAppElement("#atlas-content-modeler-fields-app");
```
nickcernis, line 8: @anthonyburchell @ChrisWiegman This is the change I mentioned from standup that fixes the failing Jest test. (It makes sure the side-effect/setAppElement is called the first time the modal is rendered, instead of as soon as the Modal file is included/run.)

```diff
diff --git a/includes/publisher/js/src/components/RelationshipModal.jsx b/includes/publisher/js/src/components/RelationshipModal.jsx
index f5fb5ff..4302d0e 100644
--- a/includes/publisher/js/src/components/RelationshipModal.jsx
+++ b/includes/publisher/js/src/components/RelationshipModal.jsx
@@ -5,8 +5,6 @@ import Modal from "react-modal";
 const { wp } = window;
 const { date, apiFetch } = wp;
 
-Modal.setAppElement("#atlas-content-modeler-fields-app");
-
 /**
  * The modal component for editing a relationship.
  *
@@ -41,6 +39,10 @@ export default function RelationshipModal({ field, isOpen, setIsOpen }) {
 		},
 	};
 
+	useEffect(() => {
+		Modal.setAppElement("#atlas-content-modeler-fields-app");
+	}, []);
+
 	/**
 	 * Handles the selection of checkbox or radio button.
 	 *

```

includes/publisher/js/src/components/Field.jsx
nickcernis: 👍 Added this as https://wpengine.atlassian.net/browse/BH-1192 and https://wpengine.atlassian.net/browse/BH-1193.

includes/publisher/scss/_relationship-modal.scss
anthonyburchell: good catch! Thank you. I've moved that in index so we should be good there. Lookin into the other classed I have added in index as well. Might need to move the buttons as well.

includes/publisher/scss/_buttons.scss
anthonyburchell: nice. got that added too.

includes/publisher/js/src/components/RelationshipModal.jsx
anthonyburchell: yep yep! That makes sense. I've changed all the strings and tested. Seems to still be working all around. Thanks!

includes/publisher/js/src/components/Field.jsx
anthonyburchell: good call! Removed. 👍

includes/publisher/js/src/components/Field.jsx
anthonyburchell: yep yep! Fixed!

includes/publisher/js/src/components/Field.jsx
anthonyburchell: ah yeah! Added!

includes/publisher/js/src/components/RelationshipModal.jsx
nickcernis: Applied in https://github.com/wpengine/atlas-content-modeler/pull/222/commits/442fad8a3ecadd00e4d940b4aa083e0cf63d39d7.

includes/publisher/js/src/components/RelationshipModal.jsx
nickcernis: Done in https://github.com/wpengine/atlas-content-modeler/pull/222/commits/2f2cba36aff8d9e498ed22d172ffc9bffef5bac8.

includes/publisher/js/src/components/RelationshipModal.jsx
nickcernis: Implemented in https://github.com/wpengine/atlas-content-modeler/pull/222/commits/9a20c830f756ca4e05aebd2d92d83a70b3067c31.

mindctrl:
I'm seeing a small CSS issue with the background color on the feedback banner dismiss button.

image

mindctrl:
Not a blocker if we want to do follow-up polish in a separate PR, but it would be nice to improve the loading here. At first it makes you think you have no posts, then a few seconds later they pop in.

A couple ideas:
- Have a loading state of some kind that lets the user know it's loading the posts
- Pre-load the data so this screen is super fast.
- Something else?

https://user-images.githubusercontent.com/1254870/130148852-ae40d94d-3ddb-4941-bf11-8e51862e7869.mov

mindctrl:
Related to above, and also could be addressed separately:

The lag between saving and the selected post showing up makes it feel like it's not working right at first.

https://user-images.githubusercontent.com/1254870/130149165-d1a9a74e-32ff-447c-ab75-d85fc3c35c9d.mov

nickcernis:
I'm working on some final tasks for this today:

- [ ] Improve modal pagination button hover/disabled states.
- [ ] Make buttons say “Link [Singular|Plural Name of Model]” instead of “Link Reference”.
- [ ] Loading states.
- [ ] Refactor to reduce component size/complexity.
- [ ] Drop-downs to remove entries.

includes/publisher/js/src/components/relationship/modal/index.jsx
```diff
@@ -0,0 +1,231 @@
...
+	return (
+		<Modal
+			isOpen={isOpen}
+			contentLabel={`Creating relationship with ${field.name}`}
+			parentSelector={() => {
+				return document.getElementById(
+					"atlas-content-modeler-fields-app"
+				);
+			}}
+			portalClassName="atlas-content-modeler-edit-model-modal-container atlas-content-modeler"
```
mindctrl, line 154: `atlas-content-modeler-edit-model-modal-container` looks like a remnant of copy paste?

nickcernis: Thanks! Fixed in https://github.com/wpengine/atlas-content-modeler/pull/222/commits/2f1b03189834e35d004b61811b5d5944b94acc79.

mindctrl:
Looking good! Confirmed all the tasks listed here are working as expected: https://github.com/wpengine/atlas-content-modeler/pull/222#issuecomment-903582161

Once we merge #247 into this branch and get the e2e tests done, I think we'll be ready to merge this into `relationship`.

nickcernis (approved):
🥳

mindctrl (approved):

PR #225, opened 2021-08-13
[BH-1183] Integrate forked library for managing relationships

apmatthews:
## Description

https://wpengine.atlassian.net/browse/BH-1183

### The Library

The 10up/wp-content-connect library provides everything we need to manage post-to-post relationships through a custom database table. I've forked the library and implemented the following changes:

1. **Stripped out all UI code.** We're building our own, so anything related to the default UI was unnecessary.
2. **Stripped out post-to-user relationships.** Our relationship fields only need to associate ACM models. The built-in WP User is not an ACM model.
3. **Updated the `add_relationship()` and `delete_relationship()` functions** to prevent relationship creation when the given post IDs do not match the relationship's "from" or "to" post type. Without this, you could have a Car -> Tire relationship object called `$carTire`, run `$carTire->add_relationship( 1, 2 )`, and the relationship would be created even if post IDs 1 and 2 were neither a Car nor a Tire.
4. **Add support for one way relationships.** By default, the library treated all relationships as bidirectional. This resulted in two DB records per relationship in order to allow queries from either side. Now, only one record will be created when the `is_bidirectional` argument is set to `false`.
5. Converted namespaces, hook names, function names, table names, and custom WP_Query vars to assume that the library is being used in Atlas Content Modeler. This should help prevent conflicts with sites or other plugins that are already running the 10up library. Thanks @nickcernis for the suggestion!

The fork's README has been updated to reflect all of these changes. Make sure to head over there and look over things as part of this review: [ACM Content Connect](https://github.com/wpengine/acm-content-connect)

### Relationship registration

This PR also defines relationships in the library's registry for any relationship fields that a user has created.

## Testing

First, you'll need to run `composer install` to pull down the package.

After that, the only observable change will be the creation of a `wp_acm_post_to_post` table in your database.

Registration of relationship fields with the library's registry is tested in PHPUnit.

**Note:** PHPUnit will spit a few benign errors during the installation phase of the tests. I've identified a root cause and will be looking into a solution.

### Fork Testing

The library came with a thorough suite of tests, which all still run successfully after the above modifications. I still need to spend some time adding tests for the new functionality that was introduced, but wanted to complete the basic integration work first.

## Dependent PRs

Should be useful for implementing the back-end save process in #222.


mindctrl:
Nice work. Good idea to validate the post type when adding/deleting a relationship.

> Add support for one way relationships. By default, the library treated all relationships as bidirectional. This resulted in two DB records per relationship in order to allow queries from either side. Now, only one record will be created when the is_bidirectional argument is set to false.

Can you explain why you did this? Can you explain the difference in a one-way and a two-way relationship in this context? If a Car <> Tire relationship exists, can you explain the mental model of why that's a one-way relationship only, vs being a two-way relationship that's just being viewed from a different perspective?

apmatthews:
> If a Car <> Tire relationship exists, can you explain the mental model of why that's a one-way relationship only, vs being a two-way relationship that's just being viewed from a different perspective?

A Car <> Tire relationship can only be two way if you know in advance what the relationship is called on a Car and what it's called on a Tire. Our current UI only collects half of that information, which means we can only create one way relationships.

Honestly, I'm not sure if we need one way relationships going forward or not. I struggle to imagine real use cases for one way relationships, but [Graph CMS has them](https://graphcms.com/docs/api-reference/schema/field-types#one-way-references). You can't convert a one way field to a two way field, so I think there's some potential for user regret. I wouldn't mind treating all relationships as two way provided we had the following:

1. A way to disable the UI on either side of the relationship to make for a cleaner publisher experience.
2. A way to toggle `show_in_graphql` and `show_in_rest` for each side of the relationship.
3. Reverse field name and slug from the user, even if they don't intend to use the reverse reference.

> Can you explain why you did this?

Aside from the challenge presented above, my understanding is that we are only focused on one way relationships at the moment. As long as we support creating one way relationships, I think it makes sense to optimize the library to avoid creating unnecessary records and potentially confusing queries (happy to explain this part more if needed, but it gets deep).

apmatthews:
Putting this back in draft to fix a bug I stumbled across today. Should be a fairly quick follow up on Monday!

nickcernis:
Thanks for all the thought that's gone into this!

> I could set up a new repository in the org and push the code there, but we would need to have permission to make it public.

I think we should try to do this, but am happy to link it to your repo to approve this PR for now and make the change later.

> After that, the only observable change will be the creation of a wp_post_to_post table in your database.

Is it safe to use the same table name in our forked version as in 10up's version? (Might we need to change the schema independently of their implementation, for example? Or could we run into issues if another plugin is using 10up's lib?)

>  By default, the library treated all relationships as bidirectional. This resulted in two DB records per relationship in order to allow queries from either side. Now, only one record will be created when the is_bidirectional argument is set to false.

Doesn't this make it harder to expose the reverse relationship in our UI? Are we confident now that we will not want to do that later? I wonder if there is value in maintaining the bidirectional relationship even though we currently expose the relationship in our UI in one direction only.

ChrisWiegman (dismissed):
The code seems to work but I think it is a requirement that this is moved under the wpengine org before we merge.

apmatthews:
> Is it safe to use the same table name in our forked version as in 10up's version? (Might we need to change the schema independently of their implementation, for example? Or could we run into issues if another plugin is using 10up's lib?)

Good question. I do think there is some risk involved in sticking with the default table name. Adding an `acm` prefix would be preferable, but it was hard to determine where to stop modifying the library. Initially I had only planned on removing the pieces of functionality we didn't need so that we could potentially merge in any upstream changes. However, considering the two additions I've made and various conversations I've had, maybe we are already moving towards complete divergence? In that case, customizing table names seems appropriate. We should also change the relevant PHP namespaces, WP hook names, and `WP_Query` args to fully address potential conflicts with third-party code that runs the unaltered library. Any concerns with this plan?

> Doesn't this make it harder to expose the reverse relationship in our UI?

Yes, but that's the choice we are giving the developer by offering one way and (eventually) two way relationships. This choice should be made in the relationship field UI during the initial creation of the field.

> Are we confident now that we will not want to do that later?

Even if we kept both DB records we still couldn't expose the reverse relationship later. The `name` that we give a relationship is different depending on whether it is one way or two way -- that `name` is stored in the DB alongside each record. Exposing a two way relationship requires knowing both the forward and reverse field information at the time of creation so that we can generate a shared `name` that acknowledges both sides of the relationship. Using the shared `name` for both forward and reverse queries is what allows the two sides to remain in sync.

> The code seems to work but I think it is a requirement that this is moved under the wpengine org before we merge.

I've created the repo. I will make it public and change the composer reference tomorrow after addressing the feedback so far.

nickcernis:
>  In that case, customizing table names seems appropriate. We should also change the relevant PHP namespaces, WP hook names, and WP_Query args to fully address potential conflicts with third-party code that runs the unaltered library. Any concerns with this plan?

That seems ok to me, given that the 10up repo mentions that they don't intend to make big changes (that we'd have to track). Might be worth discussing with the whole team in case there's a hidden cost we're missing, though.

> Even if we kept both DB records we still couldn't expose the reverse relationship later. The name that we give a relationship is different depending on whether it is one way or two way -- that name is stored in the DB alongside each record. Exposing a two way relationship requires knowing both the forward and reverse field information at the time of creation so that we can generate a shared name that acknowledges both sides of the relationship. Using the shared name for both forward and reverse queries is what allows the two sides to remain in sync.

Thanks for the explanation. If we can't generate that name without input from the user this makes sense. (Unless we can infer the slug for the other relationship from the model types?) It sounds like we're leaning towards making one-way vs two-way an explicit user choice, though, so we probably only want to set up relationships as two-way if the user has opted in to that.

atlas-content-modeler.php
```diff
@@ -51,4 +57,7 @@ function atlas_content_modeler_loader(): void {
 
 	$form_editing_experience = new \WPE\AtlasContentModeler\FormEditingExperience();
 	$form_editing_experience->bootstrap();
+
+	// Boostrap relationships library.
+	\WPE\AtlasContentModeler\ContentConnect\Plugin::instance();
```
mindctrl, line 62: This is failing in the e2e test. We need to run `composer install --no-dev` somewhere in the Circle workflow.

atlas-content-modeler.php
```diff
@@ -51,4 +57,7 @@ function atlas_content_modeler_loader(): void {
 
 	$form_editing_experience = new \WPE\AtlasContentModeler\FormEditingExperience();
 	$form_editing_experience->bootstrap();
+
+	// Boostrap relationships library.
```
mindctrl, line 61:
```suggestion
	// Bootstrap relationships library.
```

atlas-content-modeler.php
mindctrl: Composer already runs. I think this is due to the `*vendor*` pattern in `.zipignore`. https://github.com/wpengine/atlas-content-modeler/blob/4d66261de94b721b62a2a4ffc7b0710d091bf003/.zipignore#L24

apmatthews: I want to make sure that once we remove `*vendor*` from the zipignore we're not packaging dev dependencies in the zip. I could add the `--no-dev` flag to the `plugin-build-composer` job, but that would probably break the testing jobs. Do we need a second `plugin-build-composer` job for deployment?

apmatthews: All green! I did end up splitting the composer install into two jobs. One for the --no-dev build that goes into the zip, and another that installs dev dependencies. It also cut down on a few repetitive installs that were built directly into certain jobs.

Screen Shot 2021-08-18 at 12 23 02 PM

mindctrl (approved):

ChrisWiegman (approved):

PR #226, opened 2021-08-13
[NO JIRA] build: Use alias to map `@wordpress/i18n` dependency to WP global. #182

### LINKED ISSUE #182: Add alias for @wordpress/i18n dependency
We have `@wordpress/i18n` set up as a dependency here:
https://github.com/wpengine/atlas-content-modeler/blob/061c95e2eb8e8d776418f2f895b976634a706196/package.json#L26

We can keep it out of our bundle and use the one that ships with WP if we add it to the aliases here:
https://github.com/wpengine/atlas-content-modeler/blob/061c95e2eb8e8d776418f2f895b976634a706196/package.json#L59

We need to make sure the `wp-i18n` script is loading on our pages so the alias works.

mindctrl:
## Description
This instructs Parcel to map imports to the `wp` global available at runtime when the `wp-i18n` scripts is enqueued.
See: https://v2.parceljs.org/features/module-resolution#aliases

Closes #182 

Build file size differences, using `npm run build`:
`includes/publisher/dist/index.js`
before - 54k
after - 40k

`includes/developer/dist/index.js`
before - 332k
after - 320k

### DIFF
--- a/tests/integration/publisher/test-script-dependencies.php
+++ b/tests/integration/publisher/test-script-dependencies.php
@@ -1,0 +1,63 @@
+<?php
+/**
+ * Tests for script dependencies.
+ *
+ * @package AtlasContentModeler
+ */
+
+use WPE\AtlasContentModeler\FormEditingExperience;
+use function WPE\AtlasContentModeler\ContentRegistration\update_registered_content_types;
+
+/**
+ * Class TestPublisherScriptDependencies
+ */
+class TestPublisherScriptDependencies extends WP_UnitTestCase {
+
+	private $models = [
+		'recipe' => [
+			'singular' => 'Recipe',
+			'plural'   => 'Recipes',
+			'slug'     => 'recipe',
+		],
+	];
+
+	public function setUp() {
+		wp_set_current_user( 1 );
+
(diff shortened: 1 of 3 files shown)


nickcernis (dismissed):

tests/integration/publisher/test-script-dependencies.php
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	public function tearDown() {
+		parent::tearDown();
+		wp_set_current_user( null );
+		delete_option( 'atlas_content_modeler_post_types' );
+	}
+
+	public function test_wp_i18n_script_is_enqueued_on_the_publisher_view(): void {
+		self::assertTrue( wp_script_is( 'wp-i18n', 'enqueued' ) );
```
nickcernis, line 49: Nice! I didn't know about `wp_script_is()`.

nickcernis (dismissed):

nickcernis (approved):

ChrisWiegman (approved):

anthonyburchell (approved):
Nice! LGTM 👍

PR #230, opened 2021-08-17
BH-1094 Logo for plugin

matthewguywright:
## Description

Updated the svg for the plugin image.


includes/settings/settings-callbacks.php
```diff
@@ -45,6 +45,20 @@ function register_admin_menu_page(): void {
 	);
 }
 
+add_filter( 'all_plugins', __NAMESPACE__ . '\prefix_add_plugin_icon' );
+/**
+ * Add a custom plugin icon to the plugins update page (introduced in WP4.9).
+ *
+ * @param array|mixed $plugins Plugins passed in.
```
ChrisWiegman, line 52: Neither param nor return value should be mixed (see https://developer.wordpress.org/reference/hooks/all_plugins/)

includes/settings/settings-callbacks.php
```diff
@@ -45,6 +45,20 @@ function register_admin_menu_page(): void {
...
+ *
+ * @return mixed
+ */
+function prefix_add_plugin_icon( $plugins ) {
+	$plugins['plugin-directory/plugin-directory.php']['icons']['default'] = plugin_dir_url( __FILE__ ) . 'img/Atlas.svg';
```
ChrisWiegman, line 57: Are there any known ways these indexes might be missing?

ChrisWiegman, line 57: Is this plugin slug correct for our plugin?

ChrisWiegman (dismissed):
I'm afraid I have been unable to get this to work for me locally.

Screen Shot 2021-08-17 at 2 52 12 PM

includes/settings/settings-callbacks.php
mindctrl: `'plugin-directory/plugin-directory.php'` this part isn't correct. It needs to target our main plugin file.

includes/settings/settings-callbacks.php
```diff
@@ -45,6 +45,20 @@ function register_admin_menu_page(): void {
 	);
 }
 
+add_filter( 'all_plugins', __NAMESPACE__ . '\prefix_add_plugin_icon' );
```
mindctrl, line 48: I think we should target this in the update check functionality. It'll run in less places, it'll look more like what WP core does, and it'll allow us to change the logo later from the API if we want - as opposed to deploying a new logo in the plugin.

nickcernis (changes requested):
I owe @matthewguywright an apology here. It was me who added the link to the sample plugin icon update code on the Jira card, thinking it would be a helpful starting point.

It seems that code no longer works (not just due to the plugin name difference pointed out above — even when using `atlas-content-modeler/atlas-content-modeler.php` instead of `plugin-directory/plugin-directory.php` the `all_plugins` filter does not fire on the plugin update page). Sorry for not testing that code, Matt.

To set a fallback icon in the update request logic as @mindctrl suggests, this works for me:

```diff
diff --git a/includes/settings/settings-callbacks.php b/includes/settings/settings-callbacks.php
index 3e38814..f1b864a 100644
--- a/includes/settings/settings-callbacks.php
+++ b/includes/settings/settings-callbacks.php
@@ -45,20 +45,6 @@ function register_admin_menu_page(): void {
 	);
 }
 
-add_filter( 'all_plugins', __NAMESPACE__ . '\prefix_add_plugin_icon' );
-/**
- * Add a custom plugin icon to the plugins update page (introduced in WP4.9).
- *
- * @param array|mixed $plugins Plugins passed in.
- *
- * @return mixed
- */
-function prefix_add_plugin_icon( $plugins ) {
-	$plugins['plugin-directory/plugin-directory.php']['icons']['default'] = plugin_dir_url( __FILE__ ) . 'img/Atlas.svg';
-
-	return $plugins;
-}
-
 add_filter( 'parent_file', __NAMESPACE__ . '\maybe_override_submenu_file' );
 /**
  * Overrides the “submenu file” that determines which admin submenu item gains
diff --git a/includes/updates/update-functions.php b/includes/updates/update-functions.php
index 073da71..b462617 100644
--- a/includes/updates/update-functions.php
+++ b/includes/updates/update-functions.php
@@ -94,6 +94,10 @@ function atlas_content_modeler_get_remote_plugin_info() {
 			wp_remote_retrieve_body( $response )
 		);
 
+		if ( ! property_exists( $response, 'icons' ) || empty( $response->icons['default'] ) ) {
+			$response->icons['default'] = ATLAS_CONTENT_MODELER_URL . 'includes/settings/img/Atlas.svg';
+		}
+
 		set_transient( 'atlas_content_modeler_product_info', $response, HOUR_IN_SECONDS * 12 );
 	}
```

Screenshot 2021-08-18 at 11 40 45

To test after applying the above changes:

1. Temporarily change the Version header in `atlas-content-modeler.php` (being careful not to commit the change — it's just to create a version difference with the latest on the server so that the update appears): 

    ```diff
    -* Version: 0.5.0
    +* Version: 0.1.0
    ```

2. Clear your WordPress “transients” (cached data stored in the database for a fixed time). With WP-CLI in Local you can run `wp transient delete --all`, or you can use a plugin such as [Transients Manager](https://wordpress.org/plugins/transients-manager/).

3. Visit Dashboard -> Updates. You should see the plugin and icon. (Do not update from this page on your dev site; it will overwrite your local work in progress.)

matthewguywright:
@nickcernis @mindctrl @ChrisWiegman This has been updated and tested on my end. Thanks Nick. Feel free to confirm and test.

nickcernis (approved):
Thanks, @matthewguywright. The latest looks good to me!

Noting that it will be expected behavior for the update this code appears in to use the fallback icon, but that subsequent updates should show the ACM icon, just in case anyone mentions it.

Screenshot 2021-08-19 at 11 21 53

mindctrl (approved):

PR #231, opened 2021-08-17
[BH-1136] One to many relationship

ChrisWiegman:
*Note: This PR is against #222. Due to the nature of the changes I have not added independent tests nor have I solved for existing issues in that branch, yet.*

## Description

Expands the existing 1:1 relationship UI to be able to handle 1:many relationships.

This tries to maintain as much of the 1:1 code as possible and simply assumes that all entries could be an array of values. Values are saved as a comma-delimited string to the database until post-to-posts integration is complete. This maintains all existing functionality of the 1:1 relationships

https://wpengine.atlassian.net/browse/BH-1136

## Testing

At the moment, no additional tests have been added. I will work with Anthony for acceptable tests on the 1:1 relationship which then could be extended to this. Backend testing feels inappropriate at this juncture until we better know how the integration of 10up's library will affect data storage and other aspects.

## Screenshots

Screen Shot 2021-08-17 at 2 37 24 PM

Screen Shot 2021-08-17 at 2 37 38 PM

### DIFF
--- a/includes/publisher/js/src/components/Field.jsx
+++ b/includes/publisher/js/src/components/Field.jsx
@@ -95,10 +95,17 @@ function fieldMarkup(field, modelSlug, errors, validate) {
 			 */
 			function getRelatedTitles(field) {
 				const { models } = atlasContentModelerFormEditingExperience;
+				let values = field.value.split(",");
+				let query = "";
+
+				values.forEach((relationship) => {
+					let lead = query === "" ? "?" : "&";
+					query = query + `${lead}include[]=${relationship}`;
+				});
 
 				const endpoint = `/wp/v2/${
 					models[field.reference].wp_rest_base
-				}/?include[]=${field.value}`;
+				}/${query}`;
 
 				const params = {
 					path: endpoint,
@@ -129,11 +136,14 @@ function fieldMarkup(field, modelSlug, errors, validate) {
 			 * Caches those entries in the pagedEntries object, keyed by page.
 			 */
 			useEffect(() => {
-				getRelatedTitles(field).then((entries) => {
-					setRelatedContent(() => {
-						return entries;
+				if (field.value !== "") {
+					// let's not make the call if we don't have to.
+					getRelatedTitles(field).then((entries) => {
+						setRelatedContent(() => {
+							return entries;
+						});
 					});
-				});
+				}
 			}, [field.value]);
 
 			function relationshipClickHandler(
--- a/includes/publisher/js/src/components/Field.jsx
+++ b/includes/publisher/js/src/components/Field.jsx
@@ -172,7 +172,6 @@ function fieldMarkup(field, modelSlug, errors, validate) {
 										<section className="card-content">
 											<ul className="model-list">
 												<li>
-													{console.log(field.value)}
 													<div className="relation-model-card flex-wrap d-flex flex-column d-sm-flex flex-sm-row">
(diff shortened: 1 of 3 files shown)


includes/publisher/js/src/components/Field.jsx
```diff
@@ -95,10 +95,18 @@ function fieldMarkup(field, modelSlug, errors, validate) {
...
+				// Allow for 1:many calls
+				values.forEach((relationship) => {
+					let lead = query === "" ? "?" : "&";
+					query = query + `${lead}include[]=${relationship}`;
+				});
 
 				const endpoint = `/wp/v2/${
 					models[field.reference].wp_rest_base
-				}/?include[]=${field.value}`;
+				}/${query}`;
```
nickcernis, line 109: [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) is a built-in that might make this easier, but see what you think. (GH's “suggest” isn't working for this block; sorry for the diff):

```diff
diff --git a/includes/publisher/js/src/components/Field.jsx b/includes/publisher/js/src/components/Field.jsx
index 6bcc847..df748bd 100644
--- a/includes/publisher/js/src/components/Field.jsx
+++ b/includes/publisher/js/src/components/Field.jsx
@@ -95,18 +95,13 @@ function fieldMarkup(field, modelSlug, errors, validate) {
 			 */
 			function getRelatedTitles(field) {
 				const { models } = atlasContentModelerFormEditingExperience;
-				let values = field.value.split(",");
-				let query = "";
-
-				// Allow for 1:many calls
-				values.forEach((relationship) => {
-					let lead = query === "" ? "?" : "&";
-					query = query + `${lead}include[]=${relationship}`;
+				let query = new URLSearchParams({
+					include: field.value.split(","),
 				});
 
 				const endpoint = `/wp/v2/${
 					models[field.reference].wp_rest_base
-				}/${query}`;
+				}/?=${query}`;
 
 				const params = {
 					path: endpoint,
```

ChrisWiegman: Ah ha! Thanks! I had been looking for (unsuccessfully) just such a function. Fixed

includes/publisher/js/src/components/Field.jsx
```diff
@@ -125,24 +128,20 @@ function fieldMarkup(field, modelSlug, errors, validate) {
...
-			 * Caches those entries in the pagedEntries object, keyed by page.
+			 * Gets post information to display to the user outside of the modal.
 			 */
 			useEffect(() => {
-				getRelatedTitles(field).then((entries) => {
-					setRelatedContent(() => {
-						return entries;
+				if (field.value !== "") {
+					// let's not make the call if we don't have to.
+					getRelatedTitles(field).then((entries) => {
```
nickcernis, line 136: If you want to, you can simplify this to:

```js
					getRelatedTitles(field).then((entries) => {
						setRelatedContent(entries);
					});
```

(Passing a function to a `useState` callback like `setRelatedContent` is useful when we want to override an existing value. If we're just replacing state it's fine to just pass the new state instead of a function.)

That can be simplified further to the line below (`entries` is passed to `setRelatedContent` so there's no need for an anonymous function — we can just pass the function reference). No obligation to use it but feel free to if you think it's more readable:

```js
					getRelatedTitles(field).then(setRelatedContent);
```

(Not trying to code golf this, by the way — just reducing code if it can help legibility.)

PR #243, opened 2021-08-19
BH-1156 BH-1157 Meta Data

matthewguywright:
## Description
Metadata in place for ACM and Publisher.


includes/shared-assets/js/ga-analytics.js
```diff
@@ -0,0 +1,16 @@
...
+			(i[r].q = i[r].q || []).push(arguments);
+		}),
+		(i[r].l = 1 * new Date());
+	(a = s.createElement(o)), (m = s.getElementsByTagName(o)[0]);
+	a.async = 1;
+	a.src = g;
+	m.parentNode.insertBefore(a, m);
+})(window, document, "script", "//www.google-analytics.com/analytics.js", "ga");
+
+ga("create", "281125716", "auto");
```
matthewguywright, line 15: I got this from GA for ACM. Is this the correct number? 281125716

includes/publisher/class-publisher-form-editing-experience.php
```diff
@@ -127,6 +127,14 @@ public function enqueue_assets( string $hook ): void {
```
matthewguywright, line 127: @ChrisWiegman @mindctrl I don't think I have permission to do anything with the filters. I see one for excluding internal traffic but it's greyed out for me. That's for the IP Address blocking.

includes/publisher/class-publisher-form-editing-experience.php
```diff
@@ -127,6 +127,14 @@ public function enqueue_assets( string $hook ): void {
 
 		$plugin = get_plugin_data( ATLAS_CONTENT_MODELER_FILE );
 
+		wp_register_script(
```
matthewguywright, line 130: @mindctrl I think these two enqueue's can be in the shared_assets.php file right? Just in one place. That's where the banner is included. Thoughts? Then we won't have duplicate code.

includes/publisher/class-publisher-form-editing-experience.php
```diff
@@ -127,6 +127,14 @@ public function enqueue_assets( string $hook ): void {
...
+			'google-analytics',
```
mindctrl, line 131: This needs to be a name unique to our plugin.

includes/publisher/class-publisher-form-editing-experience.php
```diff
@@ -175,6 +183,7 @@ public function enqueue_assets( string $hook ): void {
 			wp_enqueue_script( 'atlas-content-modeler-feedback-banner' );
 		}
 
+		wp_enqueue_script( 'google-analytics' );
```
mindctrl, line 186: Related to above, this needs to be a name unique to our plugin.

includes/publisher/class-publisher-form-editing-experience.php
mindctrl: @matthewguywright yes you can register the script in the shared file, and then enqueue it where necessary.

mindctrl (changes requested):

includes/shared-assets/js/ga-analytics.js
mindctrl: I think you need the UA number? https://developers.google.com/analytics/devguides/collection/analyticsjs

matthewguywright: There isn't a UA number in Analytics that I can find. I followed a guide to find it and where theirs shows a UA number, ours shows the number above. Thoughts? Can you confirm? https://ryanmilani.com/sem-blog/data-analytics/quickly-find-ua-code/

ChrisWiegman (dismissed):
I can confirm pageviews are firing and appear to only be doing so in settings and publisher pages. LGTM

includes/shared-assets/wp_scripts/shared_assets.php
```diff
@@ -24,5 +24,21 @@ function enqueue_shared_assets() {
 		true
 	);
 
+	wp_enqueue_script(
+		'atlas-content-modeler-gtag',
+		'https://www.googletagmanager.com/gtag/js?id=G-S056CLLZ34',
+		[],
+		$plugin['Version'],
+		false
+	);
```
mindctrl, line 33: I think we should register the script here, like the one below it is, and then enqueue it the same where needed. Otherwise this loads on all admin pages, even when it's not needed.

mindctrl (changes requested):

mindctrl:
Note to selves: this shouldn't be merged into `main` until we have official guidance from legal.

includes/publisher/js/src/App.jsx
```diff
@@ -1,10 +1,35 @@
...
+ *
+ * Send custom event
+ * ReactGA.event({
+ *	category: "your category",
+ *	action: "your action",
+ *	label: "your label", // optional
+ *	value: 99, // optional, must be a number
+ *	nonInteraction: true, // optional, true/fals
+ *	transport: "xhr", // optional, beacon/xhr/image
+ *	});
```
nickcernis, line 24: Suggest removing ReactGA example docs, which could change (but maintaining the link to their readme).

includes/settings/js/src/App.jsx
```diff
@@ -10,7 +11,33 @@ import Taxonomies from "./components/Taxonomies";
...
+ *
+ * Send custom event
+ * ReactGA.event({
+ *	category: "your category",
+ *	action: "your action",
+ *	label: "your label", // optional
+ *	value: 99, // optional, must be a number
+ *	nonInteraction: true, // optional, true/fals
+ *	transport: "xhr", // optional, beacon/xhr/image
+ *	});
```
nickcernis, line 32: Same here — we probably don't need to inline these docs.

includes/publisher/js/src/App.jsx
```diff
@@ -1,10 +1,35 @@
...
+ */
+ReactGA.initialize(GA_ID, { gtagOptions: { anonymize_ip: true } });
+
 export default function App({ model, mode }) {
 	const isEditMode = mode === "edit";
 
+	ReactGA.send({ hitType: "pageview", page: "Publisher App Home" });
```
nickcernis, line 31: What do you think of this? (The publisher app doesn't really have a “Home”, but if the publisher app loads the user is viewing a publisher entry.)

```suggestion
	ReactGA.send({ hitType: "pageview", page: "ACM Publisher Entry" });
```

includes/settings/js/src/App.jsx
```diff
@@ -10,7 +11,19 @@ import Taxonomies from "./components/Taxonomies";
...
+/**
+ * React-GA4 Library and usage
+ * {@link https://github.com/PriceRunner/react-ga4#readme}
+ */
+ReactGA.initialize(GA_ID, { gtagOptions: { anonymize_ip: true } });
+
 export default function App() {
+	ReactGA.send({
+		hitType: "pageview",
+		page: "ACM Settings Home",
```
mindctrl, line 25: "Settings" is something I've been thinking about lately. We first started out this project by copying an existing 'settings' admin page from the headless plugin. There we had legit settings, here not so much.

As the devex side of our app has evolved to include many "pages", like models and taxonomies, I wonder if it's time to start referencing this by another name?

Think of this, if we later have an actual settings page, the page view data could get skewed if we report them as the same page. I wonder if we should make the distinction now? "ACM Models Home" or something similar, and then have "ACM Taxonomies Home" (or whatever) for that view. I think it would be helpful to distinguish between them. And then it naturally leads to other events like "Editing taxonomy", "Editing model", or similar. Thoughts?

nickcernis: "ACM Models Home" makes sense to me.

PR #247, opened 2021-08-23
[BH-1191] Wire posts to posts lib

ChrisWiegman:
## Description

This wires the acm content library to the plugin to allow for saving and editing relationships in a much more robust platform.

https://wpengine.atlassian.net/browse/BH-1191

### DIFF
--- a/includes/publisher/class-publisher-form-editing-experience.php
+++ b/includes/publisher/class-publisher-form-editing-experience.php
@@ -171,7 +171,11 @@ final class FormEditingExperience {
 		if ( ! empty( $post ) && ! empty( $model['fields'] ) ) {
 			foreach ( $model['fields'] as $key => $field ) {
 				if ( isset( $post->ID ) ) {
-					$models[ $this->current_screen_post_type ]['fields'][ $key ]['value'] = get_post_meta( $post->ID, $field['slug'], true );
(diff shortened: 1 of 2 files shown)


includes/publisher/class-publisher-form-editing-experience.php
```diff
@@ -296,13 +301,22 @@ public function save_post( int $post_id, WP_Post $post ): void {
 		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized, WordPress.Security.ValidatedSanitizedInput.MissingUnslash
 		$posted_values = $_POST['atlas-content-modeler'][ $post->post_type ];
 
+		$saved_replationships = array();
```
nickcernis, line 304: Typo in `$saved_replationships`.

ChrisWiegman: Good catch, Nick! This has been fixed.

nickcernis (approved):
This looks good. I confirmed that the relationship table is populated on save, and that relationships are moved when one-to-many relationships are unchecked.

Screenshot 2021-08-24 at 17 50 16

mindctrl (approved):
Looking good here. Confirmed the data is stored and retrieved as planned. Locally I merged this into the branch for #222 and it all appears to be working well together.

PR #249, opened 2021-08-25
Feature/mw=bh 1158

ChrisWiegman:
## Description


matthewguywright:
@ChrisWiegman @anthonyburchell @mindctrl Whoever gets to this, please pull and test to confirm it is working as expected. This is a new pattern to me so I modeled it off a previous one that @nickcernis did.

PR #250, opened 2021-08-25
[BH-1137] Allow GraphQL to access relationship fields

ChrisWiegman:
## Description

This PR wires the relationship field to GraphQL.

There is only one UX change to this. Users can no longer create a relationship within a model to itself.

https://wpengine.atlassian.net/browse/BH-1137

## Testing

Create two models and add a relationship field to one of them. Add content to the related model and then establish a relationship between the two. In GraphQL IDE query the parent field as well as the nodes for the related field to see their content.


includes/content-registration/custom-post-types-registration.php
```diff
@@ -556,20 +560,96 @@ function graphql_data_is_private( bool $is_private, string $model_name, $post, $
...
+					$post,
+					$args,
+					$context,
+					$info
+				);
+
+				$resolver->set_query_arg(
+					'post__in',
+					$relationship_ids
+				);
```
nickcernis, line 601: I think we need to check if `$relationship_ids` is empty here. I'm seeing _all_ relationship table entries returned for a relationship field that has no relationships attached to it at the moment.

Adding this above `$resolver->set_query_arg` solves that issue for me:

```
				if ( count( $relationship_ids ) === 0 ) {
					return [];
				}
```

ChrisWiegman: Good catch! Fixed.

includes/settings/js/src/components/fields/RelationshipFields.jsx
```diff
@@ -46,12 +53,17 @@ const RelationshipFields = ({ register, data, editing, watch, errors }) => {
...
 							— {__("Choose a model", "atlas-content-modeler")} —
 						</option>
-						{modelsAlphabetical.map((model) => {
-							return (
-								<option key={model.slug} value={model.slug}>
-									{model.plural}
-								</option>
-							);
+						{modelsAlphabetical.map((relModel) => {
+							if (model !== relModel.slug) {
```
nickcernis, line 57: If we skip the current model, sites with only one model won't display anything in the “Choose Model” dropdown:

Screenshot 2021-09-02 at 09 14 21

And that field is required before you can save the relationship field:

Screenshot 2021-09-02 at 09 15 32

Some options to fix that:

- Hide the relationships field if a site only has one model (which may be confusing, and makes documentation harder).
- Let people create a relationship field and set the model later (We would need to adapt the field registration logic to not register connections if `reference` is empty, and the publisher app to not display the relationships field until the reference is set. I think this is going to be confusing for a developer who has added a field but does not see it in the publisher app or in GraphQL.)
- Let people create connections to the model. This used to work in an earlier commit to this PR, because I demoed it working at the last sprint demo (songs could relate to other songs). I've yet to dig into why it no longer works, but I think allowing like-to-like relationships would be the ideal scenario. There are a lot of scenarios where it will be useful: you might want a “teammates” relationship field with a one-to-many relationship between employees and other employees, for example.

I'm working on another PR but happy to revisit this if no-one else picks it up in that time.

mindctrl (dismissed):

includes/settings/js/src/components/fields/RelationshipFields.jsx
mindctrl: @nickcernis I reverted this change. It seems to be working for me. I'm sure it's possible to trigger a recursion issue somehow, but I don't think it's very likely.

nickcernis: Thanks, @mindctrl. It's working for me too, even for entries that include themselves in a relationship field.

@ChrisWiegman It looks like we can allow self-references unless there's something we've missed that you ran into. 🥳

nickcernis (approved):
This is looking good to me!

Thanks for the hard work here, @ChrisWiegman and for the fixes, @mindctrl.

includes/settings/js/src/components/fields/RelationshipFields.jsx
ChrisWiegman: When I was testing with self-references last week, creating one would break the GraphQL output. Perhaps another change has fixed that?

nickcernis: Thanks, Chris, I think you must be right. Or, at least, I can't reproduce broken output for self-referencing models now.

ChrisWiegman: In further testing this morning I can no longer reproduce. Let's call this one done.

mindctrl (approved):

PR #253, opened 2021-08-26
MTKA-1334: Add opt-in usage tracking to the plugin.

matthewguywright:
## Description

Adds opt-in usage tracking to the plugin.

At this time, this records an event when:
- A new post is published
- A model is created, updated, or deleted
- A field is created, updated, or deleted
- The model list is viewed

Todo:
- [x] Verify events are properly recorded
- [x] Ensure code/text naming is consistent and appropriate
- [x] Tests


mindctrl (changes requested):

ChrisWiegman (changes requested):
The generic events are great. What worries me here is the disclosure of "what" they're building (see screenshots below). We may need to kill pageviews. We should never know what they call something, just that they're building something.

Screen Shot 2021-08-27 at 10 29 36
Screen Shot 2021-08-27 at 10 29 46
Screen Shot 2021-08-27 at 10 30 24

includes/settings/js/src/components/DeleteModelModal.jsx
```diff
@@ -118,6 +120,12 @@ export function DeleteModelModal({ modalIsOpen, setModalIsOpen, model }) {
 					await deleteModel(slug)
 						.then((res) => {
 							if (res.success) {
+								if (atlasContentModeler.usageTrackingEnabled) {
```
mindctrl, line 123: I don't love adding this kind of logic everywhere we want an event. An alternative is to put the logic inside the functions that send the events. However, the `usageTrackingEnabled` bit that is passed through `wp_localize_script` is done for two separate scripts (publisher and settings) and exposed through two different global variables by WP. We don't have a single shared script that is enqueued and localized in both places. Something to consider for the near future.

mindctrl:
This is ready for a preliminary review. I updated the original post with the remaining tasks.

includes/settings/js/src/components/DeleteModelModal.jsx
nickcernis: @mindctrl Perhaps something like this for now?

```diff
diff --git a/includes/shared-assets/js/analytics.js b/includes/shared-assets/js/analytics.js
index dfa17c6..7332f8b 100644
--- a/includes/shared-assets/js/analytics.js
+++ b/includes/shared-assets/js/analytics.js
@@ -7,6 +7,10 @@ import ReactGA from "react-ga4";
 
 const ANALYTICS_ID = "G-S056CLLZ34";
 
+const telemetryEnabled =
+	atlasContentModelerFormEditingExperience?.usageTrackingEnabled ||
+	atlasContentModeler?.usageTrackingEnabled;
+
 const maybeInitializeAnalytics = () => {
 	if (!ReactGA?.isInitialized) {
 		ReactGA.initialize(ANALYTICS_ID, {
@@ -17,13 +21,18 @@ const maybeInitializeAnalytics = () => {
 
 export const sendEvent = (data) => {
 	maybeInitializeAnalytics();
-	return ReactGA.event(data);
+	if (telemetryEnabled) {
+		return ReactGA.event(data);
+	}
 };
 
 export const sendPageView = (page = "") => {
 	maybeInitializeAnalytics();
-	return ReactGA.send({
-		hitType: "pageview",
-		page,
-	});
+
+	if (telemetryEnabled) {
+		return ReactGA.send({
+			hitType: "pageview",
+			page,
+		});
+	}
 };
```

includes/settings/js/src/components/Settings.jsx
```diff
@@ -0,0 +1,96 @@
+/* global atlasContentModeler */
+import React, { useState } from "react";
+import { __ } from "@wordpress/i18n";
+import { dispatch } from "@wordpress/data";
+
+export default function Settings() {
```
nickcernis, line 6: This isn't saving settings quite right for me at the moment:

- Disabled isn't active by default (i.e. when `atlas_content_modeler_usage_tracking` is empty).
- Clicking 'disabled' then refreshing shows no option as active.

This change seems to fix both issues:

```diff
diff --git a/includes/settings/js/src/components/Settings.jsx b/includes/settings/js/src/components/Settings.jsx
index bdaff70..77635d6 100644
--- a/includes/settings/js/src/components/Settings.jsx
+++ b/includes/settings/js/src/components/Settings.jsx
@@ -8,12 +8,12 @@ export default function Settings() {
 		atlasContentModeler.usageTrackingEnabled
 	);
 
-	function saveUsageTrackingSetting(event) {
+	function saveUsageTrackingSetting(value) {
 		// @todo catch save errors
 		dispatch("core").saveSite({
-			atlas_content_modeler_usage_tracking: event.target.value,
+			atlas_content_modeler_usage_tracking: value,
 		});
-		setUsageTracking(event.target.value);
+		setUsageTracking(value);
 	}
 
 	return (
@@ -51,10 +51,13 @@ export default function Settings() {
 													name="atlas-content-modeler-settings[usageTracking]"
 													value="0"
 													checked={
-														usageTracking === "0"
+														usageTracking === "0" ||
+														!usageTracking
 													}
-													onChange={
-														saveUsageTrackingSetting
+													onChange={() =>
+														saveUsageTrackingSetting(
+															"0"
+														)
 													}
 												></input>
 												{__(
@@ -70,8 +73,10 @@ export default function Settings() {
 												name="atlas-content-modeler-settings[usageTracking]"
 												value="1"
 												checked={usageTracking === "1"}
-												onChange={
-													saveUsageTrackingSetting
+												onChange={() =>
+													saveUsageTrackingSetting(
+														"1"
+													)
 												}
 											></input>
 											<label
```

includes/settings/js/src/components/Settings.jsx
```diff
@@ -0,0 +1,96 @@
...
+												htmlFor="atlas-content-modeler-settings[usageTrackingDisabled]"
+											>
+												<input
+													type="radio"
+													id="atlas-content-modeler-settings[usageTrackingDisabled]"
+													name="atlas-content-modeler-settings[usageTracking]"
+													value="0"
+													checked={
+														usageTracking === "0"
+													}
```
nickcernis, line 55: This isn't saving settings quite right for me at the moment:

- Disabled isn't active by default (i.e. when atlas_content_modeler_usage_tracking is empty).
- Clicking 'disabled' then refreshing shows no option as active.

This change seems to fix both issues:

```suggestion
													checked={
														usageTracking === "0" ||
														!usageTracking
													}
```

includes/settings/js/src/components/DeleteModelModal.jsx
```diff
@@ -6,6 +7,7 @@ import { removeSidebarMenuItem } from "../utils";
 import { getRelationships } from "../queries";
 import { showError } from "../toasts";
 import { useHistory } from "react-router-dom";
+import { sendEvent } from "../../../../shared-assets/js/analytics";
```
nickcernis, line 10: Worth using the alias here for consistency?

```suggestion
import { sendEvent } from "acm-analytics";
```

nickcernis (dismissed):
One suggestion above but otherwise looks good! Might also benefit from a changelog entry under "Unreleased"?

includes/settings/js/src/components/DeleteModelModal.jsx
mindctrl: Ah yes, I think my editor auto-added the path. Good catch.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Atlas Content Modeler Changelog
 
+## Unreleased
+### Added
+- Opt-in anonymous usage tracking to help us make Atlas Content Modeler better.
```
nickcernis, line 5: Worth adding this for clarification?

```suggestion
- Opt-in anonymous usage tracking to help us make Atlas Content Modeler better (disabled by default).
```

mindctrl: Definitely! Thanks!

nickcernis (dismissed):

nickcernis (dismissed):

nickcernis (approved):

PR #254, opened 2021-08-30
[NO JIRA] Add REST route for fetching all model schemas

mindctrl:
## Description

Adds a REST route for fetching the schemas for all models. We already have a route for fetching a single model, but not for fetching them all. We've already been asked internally about getting to this data, and this would be helpful to those folks while rounding out the API a bit more.

## Testing

See phpunit tests. Make authenticated and non-authenticated REST requests to `http://{YOUR SITE}/wp-json/wpe/atlas/content-models`.

## Screenshots
Authorized requests should result in a response similar to this:
image

Unauthorized requests should result in this response:
image

## Documentation Changes
Added WP REST API route for retrieving the schemas for all models.


ChrisWiegman (approved):

anthonyburchell (approved):
LGTM! 👍  super handy!

PR #256, opened 2021-08-31
[BH-1222]: Add relationship modal empty state and “create entry” link

nickcernis:
Improves the empty state of the relationships modal table:

- Removes the pagination if there are zero pages. (Previously, “Page 1 of 0” would appear if there were no entries.)
- Displays an empty message if there are no entries. 
- Adds a link in the modal to create a new entry if there are no entries. Clicking the link closes the modal and opens the “New [Singular]` page in a new tab. This removes a potential dead end, where the user would have to figure out how to create a related entry themselves and then refresh the original page to see those new entries (potentially losing unsaved work in the process).

## Testing

Includes an e2e test for the “create new entry” flow.

### To test manually

1. Create a “Monkey” model with a “name” text field set as the title, and a “Monkey Friends” relationship field that refers to other monkeys.
2. Add your first monkey.
3. Click the relationship field to open the relationship modal. You should see no entries. Click the link named, “Create a new Monkey”.
4. In the new tab that opens, create a new monkey by just filling its name, then publish it.
5. Close the tab to return to the original monkey entry.
6. The relationship modal should update to show the monkey you entered.

## Screenshots

| Before | After |
| - | - |
| Screenshot 2021-08-31 at 13 29 10 | Screenshot 2021-08-31 at 13 28 42 |

PR #257, opened 2021-08-31
BH-1211: Remove required option from Relationship field.

mindctrl:
## Description
Removes the "required" field option from the Relationship field.

This is temporary until we work out a better UX.

I did not remove the required logic within the publisher app components, as we'll be bringing this feature back in the near future, and everything should work fine with it there in the meantime.

https://wpengine.atlassian.net/browse/BH-1211

## Documentation Changes
Removed "Make this field required" option from the Relationship field.


mindctrl:
Closing this in favor of #256

PR #262, opened 2021-09-02
refactor: rename variables to provide more clarity

mindctrl:
## Description

This targets #250 and is _optional_. Much of this already existed in this form before #250, but I noticed it while reviewing. I didn't add these changes to that PR because it increases the review burden. We can merge this into `relationship` later if necessary.


nickcernis (approved):
Thanks, John, I think this is clearer.

PR #264, opened 2021-09-07
Add support for model-to-model relationships

mindctrl:
## Description

Adds a relationship field to support model-to-model relationships.

Users can now add one-to-one and one-to-many connections between model entries.

https://wpengine.atlassian.net/browse/BH-1131

## Screenshots

image

## Documentation Changes
Added a relationship field to support model-to-model relationships.


atlas-content-modeler.php
```diff
@@ -31,6 +31,12 @@
  * Bootstraps the plugin.
  */
 function atlas_content_modeler_loader(): void {
+	$composer_autoloader = __DIR__ . '/vendor/autoload.php';
+
+	if ( file_exists( $composer_autoloader ) ) {
+		include_once $composer_autoloader;
+	}
```
mindctrl, line 38: A couple thoughts on this...

If the file does not exist, we have a problem and the plugin will not work, but we're not dealing with the problem... we're sweeping under the rug. The plugin will fatal error on line 64 below if this file is missing.

If we're not going to deal with the problem a missing file causes here, this check doesn't help us prevent bad things from happening. We could remove the check, or prevent the plugin from loading any further.

Related: our build pipeline, in its current state, would never deploy this plugin with the missing dependencies. Even if the plugin didn't fatal error due to missing files on a regular page load, the integration and e2e tests would not pass because of the missing dependencies.

We can log this separately in Jira and make it more robust in a separate PR.

includes/publisher/class-publisher-form-editing-experience.php
```diff
@@ -331,6 +375,83 @@ public function save_post( int $post_id, WP_Post $post ): void {
...
+	 * Retrieves the related post ids
+	 *
+	 * @param int    $post_id The post ID of the parent post.
+	 * @param string $post_type The post type of the parent post.
+	 * @param string $relationship_type The post type of the relationship.
+	 * @param string $field_name The slug of the relationship saved in the DB.
+	 *
+	 * @return string A comma separated list of connected posts
+	 */
+	public function get_relationship_field( $post_id, $post_type, $relationship_type, $field_name ) {
```
mindctrl, line 414: Declare argument and return types here?

includes/publisher/class-publisher-form-editing-experience.php
```diff
@@ -331,6 +375,83 @@ public function save_post( int $post_id, WP_Post $post ): void {
...
 	}
 
+	/**
+	 * Saves relationship field data using the post-to-posts library
+	 *
+	 * @param string  $field_id The name of the field being saved.
+	 * @param WP_Post $post The ID of the post being saved.
+	 * @param string  $field_value The post IDs of the relationship's destination posts.
+	 */
+	public function save_relationship_field( $field_id, $post, $field_value ) {
```
mindctrl, line 385: Declare argument and return types here?

includes/publisher/class-publisher-form-editing-experience.php
nickcernis: Thanks, done in https://github.com/wpengine/atlas-content-modeler/pull/264/commits/7f6fd2aef9b5ffe70a521fa707b929cce2176ff2.

includes/publisher/class-publisher-form-editing-experience.php
nickcernis: Applied in https://github.com/wpengine/atlas-content-modeler/pull/264/commits/7f6fd2aef9b5ffe70a521fa707b929cce2176ff2.

atlas-content-modeler.php
nickcernis: Agreed — we could perhaps silently fail to load the plugin, deactivate it, or display a user-facing error on the plugins page if the composer autoloader is missing.

Ticketed as [BH-1226](https://wpengine.atlassian.net/browse/BH-1226).

There may also be a case to inline the dependency and remove the autoloader, since we're the only consumers and the code is already quite specific to our plugin (it uses `acm_` prefixes etc.).

ChrisWiegman (approved):

nickcernis (approved):

PR #274, opened 2021-09-15
BH-1197: import/export

matthewguywright:
## Description

Import/Export main feature branch.


includes/settings/js/src/components/Tools.jsx
```diff
@@ -0,0 +1,130 @@
...
+										"atlas-content-modeler"
+									)}
+								</h4>
+								<p className="help">
+									{__(
+										"Select a .json file containing model/field definitions to import as a content model.",
+										"atlas-content-modeler"
+									)}
+								</p>
+								<button className="button button-primary link-button">
```
mindctrl, line 98: Looks like button styles are off? You see same?

What I see:
image

Should look like:

image

includes/settings/js/src/components/Tools.jsx
```diff
@@ -0,0 +1,130 @@
...
+				if (res) {
+					const jsonData = JSON.stringify(res);
+					const blob = new Blob([jsonData], { type: "text/plain" });
+					const url = URL.createObjectURL(blob);
+					const link = document.createElement("a");
+					link.download = `${getFormattedTime()}-model-export.json`;
+					link.href = url;
+					link.click();
+
+					toast(
```
mindctrl, line 51: The toast design looks off to me. Here's what I see:

image

mindctrl (changes requested):

includes/settings/js/src/components/Tools.jsx
matthewguywright: This has been updated.

includes/settings/js/src/components/Tools.jsx
matthewguywright: Updated.

includes/settings/js/src/components/Tools.jsx
```diff
@@ -0,0 +1,85 @@
...
+				<h2>{__("Tools", "atlas-content-modeler")}</h2>
+			</section>
+			<section className="card-content">
+				<div className="row">
+					<div className="col-xs-10 col-lg-4 order-1 order-lg-0">
+						<div className="row">
+							<div className="col-xs-12">
+								<h4>
+									{__(
+										"Import Model",
```
mindctrl, line 46:
```suggestion
										"Import Models",
```

I know the mocks show singular, but it's more accurate to use plural.

includes/settings/js/src/components/Tools.jsx
```diff
@@ -0,0 +1,85 @@
...
+									)}
+								</p>
+								<button className="button dark-blue button-primary link-button">
+									{__("Select File", "atlas-content-modeler")}
+								</button>
+							</div>
+							<div className="col-xs-12 mt-4">
+								<h4>
+									{__(
+										"Export Model",
```
mindctrl, line 63:
```suggestion
										"Export Models",
```

Same as above.

includes/settings/js/src/components/Tools.jsx
```diff
@@ -0,0 +1,85 @@
...
+										"atlas-content-modeler"
+									)}
+								</h4>
+								<p className="help">
+									{__(
+										"Exporting a model will generate a .json document representing all of the existing models and fields.",
```
mindctrl, line 69:
```suggestion
										"Exporting models will generate a .json document representing all of the existing models and fields.",
```

includes/settings/js/src/components/Tools.jsx
```diff
@@ -0,0 +1,85 @@
...
+							<div className="col-xs-12">
+								<h4>
+									{__(
+										"Import Model",
+										"atlas-content-modeler"
+									)}
+								</h4>
+								<p className="help">
+									{__(
+										"Select a .json file containing model/field definitions to import as a content model.",
```
mindctrl, line 52:
```suggestion
										"Select the .json file containing model/field definitions to be imported.",
```

includes/settings/js/src/components/Tools.jsx
```diff
@@ -0,0 +1,85 @@
...
+									)}
+								</h4>
+								<p className="help">
+									{__(
+										"Exporting a model will generate a .json document representing all of the existing models and fields.",
+										"atlas-content-modeler"
+									)}
+								</p>
+								<ExportFileButton
+									fileTitle={`${getFormattedDateTime()}-model-export.json`}
```
mindctrl, line 74: I think we should include something here that identifies the file came from ACM. Something like `acm-models-export-DATETIME.json` or `atlas-content-modeler-models-export-DATETIME.json`

Also I think it helps to put the datetime at the end, so the filename "reads" more like a short sentence.

includes/settings/scss/_buttons.scss
```diff
@@ -47,6 +47,16 @@ button.tertiary {
 	color: $color-text;
 }
 
+button.dark-blue {
+	background-color: $color-blue-dark !important;
+
+	&:focus,
+	&:hover {
+		background: lighten($color-blue-dark, 10%) !important;
+		border: 1px solid lighten($color-blue-dark, 10%) !important;
```
mindctrl, line 56: Can we make it so `!important` isn't required here? These are starting to creep more and more into our styles.

mindctrl (changes requested):

includes/settings/js/src/components/Tools.jsx
nickcernis: I like `acm-models-export-DATETIME.json`.

includes/settings/js/src/components/Tools.jsx
```diff
@@ -0,0 +1,85 @@
...
+	function getFormattedDateTime() {
+		var today = new Date();
+		var y = today.getFullYear();
+		var m = today.getMonth() + 1;
+		var d = today.getDate();
+		var h = today.getHours();
+		var mi = today.getMinutes();
+		var s = today.getSeconds();
+		return m + "-" + d + "-" + y + "-" + h + "-" + mi + "-" + s;
+	}
```
nickcernis, line 32: Worth considering a standard format like [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) with year-month-day?

For example:

```suggestion
	function getFormattedDateTime() {
		return new Date().toISOString().split('.')[0].replace(/[T:]/g, '-');
	}
```

Which gives:

```
2021-09-20-09-49-38
```

It's generally easier to sort/parse filenames in this format because sorting by name also sorts by date.

nickcernis:
@matthewguywright Welcome back! Just pinging in case you hand't seen the review comments above here. (Please feel free to ignore this if you had!)

includes/settings/scss/_buttons.scss
matthewguywright: Updated

nickcernis (dismissed):
Looking good — nice work!

I explored e2e tests with Codeception to verify the downloaded file using the [Filesystem module](https://codeception.com/docs/modules/Filesystem), but I couldn't figure out a way to set a download path for Chrome that worked on Circle and locally. I'm happy to skip this for now.

nickcernis (approved):

mindctrl (approved):

ChrisWiegman (approved):
I finally got to review this this morning and could find no other issues. Well done!

PR #275, opened 2021-09-15
Prepare plugin for wp.org

mindctrl:
## Description
- [x] Remove custom plugin updater code ([BH-1230](https://wpengine.atlassian.net/browse/BH-1230))
- [x] Address review feedback ([BH-1234](https://wpengine.atlassian.net/browse/BH-1234))
- [x] Add SVN deploy to CircleCI workflow ([BH-1229](https://wpengine.atlassian.net/browse/BH-1229))

https://wpengine.atlassian.net/browse/BH-1227


nickcernis (approved):
This looks good! Two questions:

1. Do we need to add the banners from https://wpengine.atlassian.net/browse/BH-1228 to the assets folder here, or will that be a separate PR?
2. Do we need to keep the `plugin-deploy-staging` and `plugin-deploy-production` tasks if we're deploying to WP.org and serving the zip from there? (Happy to keep them now and remove in a separate PR later once users have had a chance to upgrade if that's the plan.)

mindctrl:
@nickcernis thanks for the review!

> Do we need to add the banners from https://wpengine.atlassian.net/browse/BH-1228 to the assets folder here, or will that be a separate PR?
I was planning to do that in a separate PR, but I'll do that now since we're doing the logo in the same folder already. Thanks for the reminder.

> Do we need to keep the `plugin-deploy-staging` and `plugin-deploy-production` tasks if we're deploying to WP.org and serving the zip from there? (Happy to keep them now and remove in a separate PR later once users have had a chance to upgrade if that's the plan.)

Yes, I wanted to leave them for a couple reasons.

One being what you mentioned, an easy way for people on wp-product-info to get the update without us having to do a special/separate/manual/whatever update for that. Once we deploy there without the updater code, those sites will start checking wp.org automatically and the upgrade should be seamless.

Another is that we might want to offer some kind of opt-in beta experience in the future - that's a "maybe" and not a definite, but it gives us some room to discuss it and decide whether to remove it or keep it. And staging could be useful, separate from whether or not we stop deploying to prod.

.circleci/config.yml
```diff
@@ -23,8 +23,10 @@ commands:
       - run: svn co https://plugins.svn.wordpress.org/atlas-content-modeler --depth=empty .
       - run: svn up trunk
       - run: svn up tags --depth=empty
+      - run: svn up assets
       - run: find ./trunk -not -path "./trunk" -delete
       - run: cp -r ~/project/atlas-content-modeler/. ./trunk
+      - run: mv ./trunk/assets/wporg/* ./assets
       - run: rm -rf node_modules/
```
mindctrl, line 30: I did this instead of excluding it during `cp` because I couldn't get the exclusion to work. I kept getting syntax errors similar to this:

```
shopt -s extglob && cp -r !(node_modules) ~/project/atlas-content-modeler/. ./trunk
/bin/bash: -c: line 0: syntax error near unexpected token `('
```

ChrisWiegman (commented):
Do we want to completely remove all versions of the updater code or should we only be doing so for prod versions? Product info service gives us a staging channel which might be a great benefit to users in the future. The difference is, instead of removing this code entirely, CI would need to manage it somewhat and we would need to consider the beta versioning that would allow a user to easily switch between .org and "beta" versions. Thoughts?

mindctrl:
Good thoughts. For now I think we should just go ahead and remove it all. It'll be easy to reintroduce it, or part of it, later. In either case I think we'd have to adjust the logic if we introduced a beta setting. Leaving any of it in now would require thinking about that problem now and adjusting the logic to not override wp.org, and we don't have enough info, or really even an approved plan, to make a quick decision.

We are still deploying zips to wp-product-info staging and prod, nothing changed there, so if we wanted to offer the ability to manually try a beta, we could link people to the staging endpoint download link to try it out that way.

ChrisWiegman (approved):

PR #276, opened 2021-09-17
[BH-1195] Enable reverse relationships

ChrisWiegman:
## Description

Enables reverse relationships in both publisher and in GraphQL.

https://wpengine.atlassian.net/browse/BH-1195

## Notes

There are a few differences from the mocks. First, I did not move the name field as it is not part of the field itself and would require a more significant refactor of the settings app. Second, we still only have two choices for the relationship, one-to-one and one-to-many. This allows us to enable this ticket with the existing library and will still allow for constraints such as min/max which will cover most of their intended function.

This modifies the relationship name as stored in the WordPress database. The name is used for queries and is a breaking change. The previous name was just the field name. This could cause conflicts over time as that is independent of the connection itself and multiple fields can have the same name. With reverse relationships this means that we opened the potential to return items we didn't intend to. The new name uses the field id only.

Given our current adoption I haven't done anything to update this which means existing relationships will break. I can add to this if we feel it necessary.

Finally, do notice that the reverse-slug is populated regardless of whether the two-way relationship is enabled. This is simply a UX benefit to the user and can be changed

## Testing

Appropriate PhpUnit and Codeception tests are included.

## Screenshots

screenshot-acm lndo site-2021 09 20-12_28_43

screenshot-acm lndo site-2021 09 20-12_27_52


includes/content-registration/custom-post-types-registration.php
```diff
@@ -105,7 +105,8 @@ function register_relationships( $registry ) {
 				];
 
 				try {
-					$registry->define_post_to_post( $post_type, $field['reference'], $field['slug'], $args );
+					$name = $models[ $model ]['slug'] . '-' . $field['slug'] . '-' . $field['reverseSlug'];
```
nickcernis, line 108: Just noting that this results in a PHP warning if a relationships field was created before this PR: “Undefined index: reverseSlug”.

includes/settings/js/src/components/fields/RelationshipFields.jsx
```diff
@@ -206,6 +230,205 @@ const RelationshipFields = ({ register, data, editing, watch, errors }) => {
...
+							}
+							ref={register({
+								required: true,
+								maxLength: 50,
+							})}
+							onChange={(e) =>
+								onChangeGeneratedValue(e.target.value)
+							}
+							readOnly={editing}
+						/>
```
nickcernis, line 365: This field currently gets cleared if I:

1. Add a new relationship field.
2. Immediately click “Configure Reverse Reference” (without setting other fields such as, ”Name” and “API Identifier”).
3. Choose a model from the “Model To Reference” field.

I'm looking into the cause…

https://user-images.githubusercontent.com/647669/134879676-b614db20-f9c0-4235-853d-d189167fc64d.mp4

includes/settings/js/src/components/fields/RelationshipFields.jsx
```diff
@@ -206,6 +230,205 @@ const RelationshipFields = ({ register, data, editing, watch, errors }) => {
...
+							placeholder="Reverse Display Name"
+							type="text"
+							ref={register({
+								required: true,
+								maxLength: 50,
+							})}
+							onChange={(e) => {
+								setInputGeneratorSourceValue(e.target.value);
+								setNameCount(e.target.value.length);
+								clearErrors("reverseSlug");
```
nickcernis, line 297: `clearErrors` is not currently available here, so typing in the reverseName field generates errors at the moment. You can fix this by adding `clearErrors` to the `RelationshipFields` props at the top of `RelationshipFields.jsx` (it's already passed from `Form.jsx`):

```diff
const RelationshipFields = ({
	register,
	data,
	editing,
	watch,
	errors,
	setValue,
	model,
+	clearErrors,
}) => {
```

includes/content-registration/custom-post-types-registration.php
ChrisWiegman: Good catch. I thought I had caught all those.

ChrisWiegman: Fix pushed

includes/settings/js/src/components/fields/RelationshipFields.jsx
ChrisWiegman: Fix pushed

includes/settings/js/src/components/fields/RelationshipFields.jsx
```diff
@@ -206,6 +231,205 @@ const RelationshipFields = ({ register, data, editing, watch, errors }) => {
...
+					/>
+					<input
+						id="reverseSlug"
+						name="reverseSlug"
+						type="hidden"
+						ref={register}
+						defaultValue={
+							data?.reverseSlug
+								? data.reverseSlug
+								: toValidApiId(models[model].plural)
```
nickcernis, line 428: Is `toValidApiId(models[model].plural)` the correct fallback value here if the reverse reference section is not open? Or should it be an empty string (because no value has been set)?

includes/settings/js/src/components/fields/RelationshipFields.jsx
```diff
@@ -206,6 +231,205 @@ const RelationshipFields = ({ register, data, editing, watch, errors }) => {
...
+									: toValidApiId(models[model].plural)
+							}
+							ref={register({
+								required: true,
+								maxLength: 50,
+							})}
+							onChange={(e) =>
+								onChangeGeneratedValue(e.target.value)
+							}
+							readOnly={editing}
```
nickcernis, line 365: `readOnly={editing}` here means that I won't be able to set the `reverseSlug` if:

1. I create a relationship field but do not set “Configure Reverse Reference”.
2. I edit the relationship field.
3. I tick “Configure Reverse Reference”. At this point the reverseSlug is visible but not editable.

It might be that this is expected behavior, but just flagging in case we want to think about it some more.

ChrisWiegman: This is expected behavior. This is the big trade-off I was worried about. TBH, in graphQL this will actually make sense

includes/settings/js/src/components/fields/RelationshipFields.jsx
ChrisWiegman: Upon further review, yes, this is the intention as we need the slug here in a similar fashion as we talk about in your next comment.

includes/settings/js/src/components/fields/RelationshipFields.jsx
nickcernis: I _think_ I have a fix for this in https://github.com/wpengine/atlas-content-modeler/pull/276/commits/2fcbb981fb02bb5d9b7b80c15a6661d13cd47dcb and https://github.com/wpengine/atlas-content-modeler/pull/276/commits/c6b4ed99f9b2db7bd5406d701d4bbd889e162ea8, but will do some more testing in the morning.

nickcernis (dismissed):

nickcernis (approved):

nickcernis (changes requested):
Approved but then noticed failed tests…

nickcernis (approved):

mindctrl (approved):

PR #282, opened 2021-09-22
FIX: Use a more appropriate name for the relationship

ChrisWiegman:
## Description

Uses a more appropriate name to define the relationship. This is a split off of #276 in order to keep this change separate.

### DIFF
--- a/includes/content-registration/custom-post-types-registration.php
+++ b/includes/content-registration/custom-post-types-registration.php
@@ -105,7 +105,8 @@ function register_relationships( $registry ) {
 				];
 
 				try {
-					$registry->define_post_to_post( $post_type, $field['reference'], $field['slug'], $args );
+					$name = $post_type . '-' . $field['reference'];
+					$registry->define_post_to_post( $post_type, $field['reference'], $name, $args );
 				} catch ( \Exception $e ) {
 					/**
 					 * Either the relationship already exists,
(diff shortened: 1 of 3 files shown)


mindctrl:
This PR makes it so multiple relationships from Object1 to Object2 are not possible.

I have two Car <> Tire relationships. A one-to-one and a one-to-many.

image

image

In this scenario, when I make connections on these, one overwrites the other.

image

Note the one-to-one was overwritten.

In the database it looks like this:
image

apmatthews:
Could you outline some steps to reproduce the issue you encountered with using field slug as the relationship name? Was it already a problem in one way relationships or does it only happen when trying to implement two way?

Assuming we do need to do some work on naming, there are several approaches we could take:

**Combine the related post type names, as this PR does.**

Doesn't work for the reason outlined in @mindctrl's example above.

**Combine the related field slugs.**

I favor this approach because I think it gives us all the information we need. One way relationships would just use the regular field slug (as it is coded today). Two way relationships would combine the forward field slug with the reverse field slug in alphabetical order (to ensure we get the same name from both sides).

**Combine related field slugs AND post type names.**

There may not be any harm in this approach, but I think it's unnecessary. We already have the [post type names included in the relationship object](https://github.com/wpengine/acm-content-connect/blob/a3ef265b77dd5cb3d28f4f7a4a47689fc560ac98/includes/Relationships/PostToPost.php#L35-L36) stored in the ContentConnect registry. Using `$relationship->get_related_object_ids( $post_id )` already restricts results by post type [in the WHERE clause](https://github.com/wpengine/acm-content-connect/blob/a3ef265b77dd5cb3d28f4f7a4a47689fc560ac98/includes/Relationships/PostToPost.php#L73-L75) of the SQL query. Adding the post type names to the relationship name seems redundant, but that's where having some concrete examples of how the field slug approach falls short would be helpful.

ChrisWiegman:
@apmatthews, I like your second approach and I think that's the right approach. I also think that we should, given that, role this change back into BH-1195 (also part of Phase I) and include the forward and reverse slugs whether we display the reverse relationship or not (long term I'm starting to think we should always configure the reverse anyway). This ensures that we don't break anything again after we complete phase 1. Thoughts?

mindctrl:
@ChrisWiegman What do you think about doing the change here so that this and 1195 are easier to review?

PR #285, opened 2021-09-25
[BH-1257] Ensure "many to many" or "many-to-one" are only existing cardinality options

ChrisWiegman:
## Description

Removes all invalid cardinality options enforcing either "many to many" or "many to one" options as the only options until we can better implement cardinality in the next release. 

https://wpengine.atlassian.net/browse/BH-1257

## Testing

No additional testing was needed.

## Screenshots

Screen Shot 2021-09-28 at 09 45 33


apmatthews:
I think we need to consider how this impacts the publisher UI. Before this PR, publishers would see either radio buttons or checkboxes depending on whether the relationship is configured as one-to-one (radio buttons) or one-to-many (checkboxes). While it's technically not a "breaking" change, it still feels wrong to let checkboxes slip through where there were radio buttons before. Cardinality may not have been accurate across models, but that setting still essentially worked as a "max of one relation per entry" flag.

Thinking through the publisher UI side has made me realize that `one-to-one` and `one-to-many` is actually still accurate for one-way relationships. For example, no matter how many Unicycles are assigned "Tire A", one Unicycle will always have one Tire. The problem we are trying to solve with this PR is that we can't guarantee that one Tire will always have one Unicycle, but you can't query from the Tire side anyway because it's a one-way relationship!  In other words, our current cardinality is only a problem when you expose two-way relationships as an option.

If we moved the checkbox for "Configure reverse relationship" up higher in the UI so that it's one of your first choices, we could use that choice to change the options available for "Connections" (cardinality). Choosing NOT to configure the reverse relationship gives you "One to One" and "One to Many" as your only cardinality options. Choosing TO configure the reverse relationship gives you "Many to Many" as the only cardinality option.

One benefit to this approach is that it doesn't require any migration of existing, one-way relationship fields.

As a final point to consider, I think we will have to derive cardinality options from "Configure reverse relationship" eventually because I don't think "Many to Many" or "Many to One" should ever be options for one-way relationships. Saying that "many Cars can have many Tires" is no different than saying "one Car can have many Tires" if you're only ever querying from a single Car entry (i.e. give me all Tires related to Car with ID 9). Likewise, saying that "many Unicycles can have one Tire" is no different than saying "One Unicycle can have one Tire". The left side of the cardinality phrase is irrelevant in a one-way relationship. It's always "One".

ChrisWiegman:
I think we're missing what cardinality is a bit. A relationship, by definition, is never truly one-way as that would imply it is one-sided, a piece of content connecting to nothing. It is always connecting two pieces of content regardless of whether we artificially limit the side from which we can query the relationship as we are proposing in our content models. By this we are still, even with our one-way view, implementing a two-way relationship.

In addition, relationships are connecting posts in a set of posts or data. 

Taken together, these bring us back to our original cardinality choices of 1:1, 1:many, many:1 and many:many. What we've implemented here, regardless of how we limit an individual models ability to connect within a set, is still a many to many relationship in that many items from one set can connect to many items of the other set.  To limit this we need to think of the set as a whole. For example, for a one-to-one relationship a unicycle must have its _own_ tire. The tire cannot be used by another unicycle nor can the same unicycle use more than one tire. Our current implementation only artificially blocks the latter and still allows multiple unicycles to share tires across the set thereby breaking the one-to-one cardinality. 

While our current hiding scheme obscures this, the end result is the same. In the end a content editor will be able to connect many items in the parent model to one item in the reference model. This is, at best, a many-to-one relationship. While John is right and we don't want an inventory system, think of the many use cases where this could still be a problem such as, for instance, a person and a driver's license. One person has their own license and, if a content editor can accidentally assign to people to the same license it will be hard to catch and could result in other problems for the site owner.

Now let's look at our existing implementation of one-to-many which is, in practice, an explicitly many-to-many relationship when viewed in terms of the sets of data in which we establish the relationship. Here I think our tire example works well. We tell the content editor that one car can select any 4 tires. Then we allow those same 4 tires to be selected for any other car.

Both of our existing implementations are, in fact, valid cardinalities but they are faux many-to-one (as the only limitation is the UI) and many-to-many relationships as opposed to one-to-one and one-to-many relationships. As we implement proper relationships throughout we need to consider the proper academic definitions of these terms, relationship and cardinality, in our implementation. In all cases a "one way relationship" is still just a UX restriction in which we're hiding one side of the relationship as the relationship still has two sides. I'll concede the library has an interesting way of handling this but it is also not correct in its establishment of a one-way relationship as, in the case of the library, a one way relationship is simply a simplification of the data retrieval to only access the relationship from one side. It's still a two way relationship, we're just only restricting our queries to query 1/2 of it.

I strongly believe we should be addressing this fact now, while we have so few users, and that we should be moving to a proper implementation of cardinality going forward. This implementation includes the acknowledgment that all relationships have two sides without exception and our only implementation of one-way relationships is strictly  a UX constraint hiding access to the relationship from one side of it. Taking this approach allows us to standardize our relationship data and control access to one side or the other of the relationship, if needed, by any other UX implementation such as access control or something else.

As for the existing publisher implementation, I also continue to propose, with such a limited user base, that we continue to push for reclassifying both our misnamed cardinalities as simply many-to-many relationships as this serves to emphasize the breaking changes as well as simplifies how we implement the relationships going forward. Remember, we're already breaking existing relationships simply by implementing a naming scheme that will allow for the reverse relationships.

All this said, I want to also acknowledge that the choice of a single cardinality, for now, is based on simplifying the initial, working, offering with the least restrictive option possible. Given that, I would be open to reconfiguring this ticket for both the many to one and many to many options as opposed to the single many-to-many option. This would be slightly easier as we've moved our upgrade script into this release and would still accurately reflect the mathematical reality of the relationships that we're building. In either case, after the release of phase 1, users will have to re-connect all their relationships in existing fields anyway so this might not be a bad way to take that.

What I want to avoid is the breaking of the actual paradigm we're implementing here. Any way we cut it we're still connecting two pieces of content in a relationship which is, by definition, two way regardless of the limited storage and query visibility of our library. Over time I think we'll find that this approach is the correct one for our users as it allows us the freedom to always be able to see the full relationship while still allowing the user to impose restrictions on viewing the data from one side or the other.

ChrisWiegman:
I’ve been trying to find a better way to diagram this and stumbled upon this post which does a fairly good job of explaining data cardinality. https://www.complexsql.com/database-cardinality-database-cardinality-with-real-examples/

ChrisWiegman:
After my comment yesterday, I've refactored this with both the many-to-many AND the many-to-one relationships to reflect the actual cardinality between the two datasets.

nickcernis (approved):
I found the same thing as @apmatthews did in Slack where pre-existing relationship fields fail to show with either option selected, but otherwise this looks good! Will be good to re-test with GraphQL queries in #276 when this is merged.

(Noting we accepted breaking changes or will migrate old field cardinality type with the migration in [BH-1266](https://wpengine.atlassian.net/browse/BH-1266).)

ChrisWiegman:
@nickcernis and @apmatthews, I've found it now as well. I think the better approach to this is to move BH-1266 back into this release. This is the upgrade script. While we could attack this problem in other ways, I believe this is, in fact, what we really want anyway.

apmatthews (approved):

PR #297, opened 2021-10-04
[BH-1254] Ensure field slugs and reverse slugs are unique in GraphQL

ChrisWiegman:
## Description

Checks new and edited fields to ensure a slug conflict does not occur when a reverse-relationship is connected. When a duplicate slug is found an error is shown on the appropriate field.

https://wpengine.atlassian.net/browse/BH-1254

## Testing

Appropriate unit tests have been created to determine when a collision should occur and check for the appropriate error.

## Notes

I've done what I believe is extensive testing of new and edited fields to ensure the error is present when needed and only when needed. Please verify the logic in your own testing as, I fully admit, the determination of the error condition was far trickier than I had anticipated.

### DIFF
--- a/includes/rest-api/fields.php
+++ b/includes/rest-api/fields.php
@@ -125,7 +125,9 @@ function content_model_field_exists( string $slug, string $id, string $model_id 
 		foreach ( $type['fields'] as $field ) {
 			if ( 'relationship' === $field['type'] &&
 				$model_id === $field['reference'] &&
+				isset( $field['enableReverse'] ) &&
 				true === $field['enableReverse'] &&
+				isset( $field['reverseSlug'] ) &&
 				$slug === $field['reverseSlug']
 			) {
 				return true;
--- a/includes/rest-api/fields.php
+++ b/includes/rest-api/fields.php
@@ -100,10 +100,13 @@ function cleanup_detached_relationship_references( string $slug ) {
  *
  * @param string $slug  The current field slug.
  * @param string $id    The current field id.
- * @param array  $model The content model to check for duplicate slugs.
+ * @param string $model_id The id of the content model to check for duplicate slugs.
  * @return bool
  */
-function content_model_field_exists( string $slug, string $id, array $model ): bool {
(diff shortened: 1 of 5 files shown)


nickcernis (commented):
I still need to do some more testing, but this is looking good so far, @ChrisWiegman!

Do you see this formatting issue for the error message (Chrome with parcel cache and browser cache cleared), or is it me?

Screenshot 2021-10-05 at 17 44 44

includes/rest-api/routes/content-model-field.php
```diff
@@ -147,6 +148,29 @@ function dispatch_update_content_model_field( WP_REST_Request $request ) {
...
+		isset( $params['enableReverse'] ) &&
+		true === $params['enableReverse'] &&
+		content_model_field_exists(
+			$params['reverseSlug'],
+			$params['id'],
+			$params['reference']
+		)
+	) {
+			return new WP_Error(
+				'wpe_duplicate_field_reverse_slug',
```
mindctrl, line 164: Most of our WP_Error codes are prefixed with `atlas_content_modeler_` or `atlas-content-modeler-`

ChrisWiegman:
@nickcernis, I've adjusted the message to bring it to one line as I think that looks better than anything I can figure out with CSS.

Screen Shot 2021-10-06 at 12 41 31

includes/rest-api/fields.php
```diff
@@ -100,21 +100,36 @@ function cleanup_detached_relationship_references( string $slug ) {
...
-		if ( $field['slug'] === $slug ) {
-			return true;
+
+		foreach ( $model['fields'] as $field ) {
+			if ( $current_model === $model_id && $field['id'] === $id ) {
+				continue;
+			}
+
+			if ( $model_id === $current_model &&
+			$field['slug'] === $slug ) {
```
mindctrl, line 120: Seems like this should be idented, but our linter isn't complaining so 🤷

mindctrl (approved):

nickcernis (approved):

PR #308, opened 2021-10-05
chore: Add documentation and download links

mindctrl:
* Adds a link to the new docs site.
* Updates the download link to use the wp.org package.
* Updates the minimum required WP version to match what's in the plugin header.


anthonyburchell (approved):
Yep yep! It's alive!

nickcernis (approved):

PR #316, opened 2021-10-08
BH-1189: author added to support with tests

matthewguywright:
## Description

Add 'author' to support in graphql.


mindctrl (changes requested):
A few formatting things.

Also a test failed, but appears to be unrelated to what you've done here. Let's see if they pass after you make another commit here.

ChrisWiegman (dismissed):
Other than the issues John pointed out, this works fine for me. Approving while you solve the issues he has pointed out.

mindctrl (approved):

ChrisWiegman (approved):

PR #323, opened 2021-10-12
BH-1184: add description to fields

matthewguywright:
## Description
Add description to fields in settings side and display on publisher side.


nickcernis (changes requested):
I think we should remove the original description textarea from the Relationship field. It now has two descriptions with the changes from this PR.

The related `descriptionMaxLength` and `descriptionCount` logic for that field could also be removed from `includes/settings/js/src/components/fields/RelationshipFields.jsx`.

Screenshot 2021-10-13 at 11 42 10

nickcernis (changes requested):

mindctrl (changes requested):

nickcernis (approved):

mindctrl (approved):
I think the description text is supposed to render with a color of `#405E6A`, if I'm reading the Figma file correctly, and these are not doing that. Otherwise this looks good.

PR #329, opened 2021-10-19
[BH-1263] Ensure p2p library respects cardinality

ChrisWiegman:
## Description

Ensures the posts-to-posts library respects cardinality by failing to add new relationships that would violate cardinality by throwing a "false" from the add_relationships menu. This can then be worked into our front-end in BH-1264.

https://wpengine.atlassian.net/browse/BH-1263

## Testing

New unit tests have been added to check each of the cardinality restrictions. Please pay close attention to these in review. I think they're right but, after a while, such logic becomes circular and bugs are too easy to miss.

## Technical notes

There isn't much we can do just with this until we finish BH-1262. This simply applies a method for BH-1262 to use. Note this assumes a `cardinality` param on any relationships object to function.

### DIFF
--- a/includes/content-connect/includes/Relationships/PostToPost.php
+++ b/includes/content-connect/includes/Relationships/PostToPost.php
@@ -321,21 +321,19 @@ class PostToPost extends Relationship {
 
 		switch ( $this->cardinality ) {
 			case 'one-to-one':
-				foreach ( $forward_relationships as $relationship ) {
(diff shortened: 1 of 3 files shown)


matthewguywright (approved):

nickcernis (approved):
Thanks, @ChrisWiegman!

I ended up adding the same cardinality class property for my PR https://github.com/wpengine/atlas-content-modeler/pull/333, but happy to merge this first and take care of conflicts in my PR.

PR #330, opened 2021-10-19
chore: add prefer-stable to Composer config

mindctrl:
## Description
Addresses PHP version compatibility issues with some Composer dependencies.

The following error can happen in PHP 7.x environments:

`Parse error: syntax error, unexpected 'static' (T_STATIC) in /home/circleci/project/atlas-content-modeler/vendor/symfony/string/UnicodeString.php on line 44`

This happens because the `minimum-stability` was set to `dev` [here](https://github.com/wpengine/atlas-content-modeler/commit/9e5a081dc3f9cec69933fd62e13df7a1cf9280ed#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R5).

Sometimes the 'dev' version is pinned to PHP 8, which causes the above error on PHP 7 environments.

## Testing

I tested this locally, and the test suite we have now should pass with this change.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,6 +2,7 @@
     "name": "wpengine/atlas-content-modeler",
     "description": "Plugin for content modeling in WordPress.",
     "type": "project",
+    "prefer-stable": true,
     "minimum-stability": "dev",
     "require-dev": {
         "codeception/module-asserts": "^1.3.1",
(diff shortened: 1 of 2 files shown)


ChrisWiegman (changes requested):
Can you run an appropriate "composer update" as part of this to ensure it doesn't break anything?

ChrisWiegman (approved):

nickcernis (approved):

PR #332, opened 2021-10-19
[BH-1219] Ensure max ID size matches model name

ChrisWiegman:
## Description

This modifies the formatter we use for the slug to limit output to 20 characters.

https://wpengine.atlassian.net/browse/BH-1219

## Testing

No additional tests were necessary however the expected values of existing tests were modified to match the expected output.

### DIFF
--- a/includes/rest-api/models.php
+++ b/includes/rest-api/models.php
@@ -106,7 +106,7 @@ function create_models( array $models ) {
 function get_model_args( string $post_type_slug, array $args ) {
 	$post_type_slug = sanitize_key( $post_type_slug );
 
-	if ( empty( $post_type_slug ) || strlen( $post_type_slug ) > 50 ) {
+	if ( empty( $post_type_slug ) || strlen( $post_type_slug ) > 20 ) {
 		return new WP_Error(
 			'atlas_content_modeler_invalid_id',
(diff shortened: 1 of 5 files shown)


nickcernis (changes requested):
I don't think we can do this because of the 20 character limit in `register_post_type`:

https://github.com/WordPress/WordPress/blob/8447683474ebdee773fc4fca7b3e44dea45060f4/wp-includes/post.php#L1515-L1518

```php
	if ( empty( $post_type ) || strlen( $post_type ) > 20 ) {
		_doing_it_wrong( __FUNCTION__, __( 'Post type names must be between 1 and 20 characters in length.' ), '4.2.0' );
		return new WP_Error( 'post_type_length_invalid', __( 'Post type names must be between 1 and 20 characters in length.' ) );
	}
```

We should probably add a comment in our code to that effect, because I also tried to bump it to 50 at one point before @mindctrl stopped me. :-)

nickcernis (approved):

ChrisWiegman:
@nickcernis, good catch on the taxonomy issue. I hadn't realized we were reusing the function there. I'm going to approve this myself after testing your changes to it. Once we have @mindctrl or someone else, I think we'll be in good shape. Thanks!

mindctrl (approved):

PR #334, opened 2021-10-21
[BH-1294] Ensure we're properly accounting for underscores and dashes in names

ChrisWiegman:
## Description

Previously GraphiQL links were broken when a model used a dash or underscore in the model names as we weren't properly accounting for the characters to generate proper GraphQL names (no camelcase). This fixes that by converting those characters to better match the GraphQLi schema.

See the screenshots below to see how sensitive this is in practice.

https://wpengine.atlassian.net/browse/BH-1294

Screen Shot 2021-10-21 at 15 10 57
Screen Shot 2021-10-21 at 15 11 33
Screen Shot 2021-10-21 at 15 11 49
Screen Shot 2021-10-21 at 15 12 11

### DIFF
--- a/includes/settings/js/src/components/fields/Form.jsx
+++ b/includes/settings/js/src/components/fields/Form.jsx
@@ -1,7 +1,7 @@
 import React, { useState, useContext, useRef, useEffect } from "react";
 import { useForm } from "react-hook-form";
 import Modal from "react-modal";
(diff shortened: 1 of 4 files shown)


nickcernis (approved):
Looking good to me. Great work tracking this down.

mindctrl (approved):

PR #337, opened 2021-10-27
MTKA-1265: Make relationship fields with back references editable from the reverse side

matthewguywright:
## Description
Allows the display of reverse relationships on the publisher entry screen.

PR #338, opened 2021-10-28
[MTKA-1264] Filter available posts when choosing related entries

nickcernis:
### Developer app changes
- Adds one-to-one and one-to-many cardinality types to the relationship field.
- Makes one-to-one the default cardinality type. (Just because it appears first in the list. We could change the order in a separate PR if needed.)

### Publisher app changes
- Makes entries that are already linked to another entry unselectable for one-to-one and one-to-many relationship fields.
- Displays a tooltip explaining why entries are not selectable.
- The tooltip includes a link to visit the related post in a new window, unlink the entry and save the post, return to your current post, and link the entry that was previously not available if needed. (This is a happy side-effect of the work we did in https://github.com/wpengine/atlas-content-modeler/pull/256 to automatically refresh the relationship modal when you revisit the page.)

https://wpengine.atlassian.net/browse/MTKA-1264

## Testing

- Includes a WP unit test for the presence of the new `acm_related_posts` property in REST responses. This is used to check if posts are related to the current post via a specific relationship field.
- Includes an e2e test to check unselectable items appear with an info button and tooltip.

#### Setup
Import these models via JSON (you will need to remove the .txt extension):

[acm-models-export-2021-10-29-13-55-40.json.txt](https://github.com/wpengine/atlas-content-modeler/files/7442454/acm-models-export-2021-10-29-13-55-40.json.txt)

Or create them manually:

1. Create a Cat model with a Text field.
2. Create a Dog model with a Text field, a one-to-one relationship field pointing to Cat, and a one-to-many relationship field pointing to Cat.

#### Testing
1. Create two cats.
2. Create one dog pointing to the first cat in both the one-to-one and one-to-many relationship fields.
3. Add a second dog, and try to point either field to the first cat. You should see the red icon and an explanation about why you can't select the first cat.

## Screenshots

Screenshot 2021-10-29 at 15 57 06

Screenshot 2021-10-29 at 15 57 09

## Documentation Changes

Documenting cardinality is a separate story at [MTKA-1267](https://wpengine.atlassian.net/browse/MTKA-1267).

## Dependant PRs

Not a dependent, but we will need a new PR to account for reverse relationships once those can be edited from the other side. https://github.com/wpengine/atlas-content-modeler/pull/337

## Notes
Just recording the expected behavior and logic for the restrictions as discussed in Slack:

- **One-to-one field**: dim rows that are already connected to another entry.
- **One-to-many field**: dim rows that are already connected to another entry. (If you allow a connection from “A” to a “B” when B is already connected to something else, you have broken one-to-many cardinality: B is no longer exclusively linked to A via one connection, so it becomes a many-to-many relationship instead.)
- **Many-to-one field**: any entry should be selectable, even if it’s connected to other entries. (The entry on the “To” side can be connected to many entries, so we don’t need to check for restrictions.)
- **Many-to-many field**: any entry should be selectable, even if it’s connected to other entries.

When we implement reverse relationships, we’ll also need to impose restrictions when editing from the reverse side with the cardinality inverted:

- If you’re editing a **One-to-one from the B side**, you can only change to other As that aren’t connected to Bs.
- If you’re editing a **Many-to-one from the B side**, you can only select new As that don’t already have a connection with another B.


ChrisWiegman (approved):
Thank you for the complete tests! This all seems to be working well for me.

tests/acceptance/CreateRelationshipFieldEntryCest.php
```diff
@@ -117,4 +99,81 @@ public function i_can_create_a_new_entry_from_an_empty_relationships_modal_table
...
+	 */
+	public function i_cannot_choose_an_entry_that_is_already_linked_to_another_entry( AcceptanceTester $i ) {
+		// Add a one-to-many relationship field.
+		$i->click( Locator::lastElement( '.add-item' ) );
+		$i->click( 'Relationship', '.field-buttons' );
+		$i->wait( 1 );
+		$i->fillField( [ 'name' => 'name' ], 'onetomany' );
+		$i->selectOption( '#reference', 'Companies' );
+		$i->click( 'input#one-to-many' );
+		$i->click( '.open-field button.primary' );
```
matthewguywright, line 117: Awesome tests!

matthewguywright (approved):

PR #328, opened 2021-10-18
[BH-1123] Prevent use of reserved slugs during field creation

nickcernis:
## Description

Fixes #312 by preventing the use of field slugs with reserved keywords. Works by:

• Using a list of reserved field names used by GraphQL for the default WP Post type (client-side and server-side) — see `includes/settings/reserved-field-slugs.php`.
• Using GraphQL introspection to identify additional fields registered at runtime if WPGraphQL is available (client-side) — see the new `useReservedSlugs` custom hook in `includes/settings/js/src/hooks/useReservedSlugs.js`.

https://wpengine.atlassian.net/browse/BH-1123

## Testing

• Includes unit test to check that fields created via the REST endpoint cannot use default reserved field names.
• Includes e2e test for reserved field names registered at runtime.

### To test reserved default words manually

1. Click '+' to add a new field.
2. Give the field a name of 'id'.
3. Tab to the API Identifier field.

You should see a message that the 'id' slug is reserved or in use.

### To test reserved fields registered by another plugin at runtime

The PR will prevent WPGraphQL field names such as 'seo' from being used as field names if another plugin registered them with WPGraphQL:

1. Install and activate _Yoast SEO_ and the _WPGraphQL Yoast SEO Addon_: `wp plugin install --activate wordpress-seo add-wpgraphql-seo`
2. Click '+' to add a new field.
3. Give the field a name of 'seo' (this is a field name registered with WPGraphQL by the WPGraphQL Yoast SEO Addon at run-time.)
4. Tab to the API Identifier field.

You should see a message that the 'seo' slug is reserved or in use.

## Screenshots

https://user-images.githubusercontent.com/647669/137753474-f4cdcbf9-784d-4d75-9cf5-5194128957e0.mp4

## Documentation Changes

n/a

## Dependant PRs

n/a

PR #340, opened 2021-11-02
NOJIRA: Prevent reverse reference slug conflicting with itself

nickcernis:
## Description

Adjusts handling of the relationship field reverse reference slug collision check to:

-  Prevent an issue @mindctrl reported in Slack where a relationship field with a reverse reference slug could not be updated, due to its reverse reference slug colliding with itself.
- Separate error handling logic and error messages for the two types of collision. (Previously the same error was reported for both of the cases below, which is fine, but I think it's easier to reason about the code and the errors when they're separate.)
    -  `atlas_content_modeler_reverse_slug_conflict` – the reverse slug conflicts with a field slug in the reference model (the reverse slug is added to the reference model as a field, so it can't clash with existing field slugs there).
    -  `atlas_content_modeler_reverse_slug_in_use` – the reverse slug is in use by a different relationship field's reverse slug from the current model that shares the same reference (two separate relationship fields can't be added to the same reference model if they share a back-reference name).

## Testing

- Adapts existing REST endpoint unit test to check for new error types.
- Adds e2e test to confirm a relationship field with a reverse reference can be updated without conflicting with itself.
- Adds new e2e tests to check for reverse slug conflict messages for the two error cases above.

#### Test 1

From John's steps to reproduce:

- Create new model
- Add relationship field
- Give it a name
- Select a Model to Reference
- Check Configure Reverse Reference
- Give that a name and Reverse API Identifier
- Save field
- Edit field, make no changes... or add a description if you like.
- Click Update

You should no longer see the error visible on `main`: "A field in the related model has the same API identifier"

#### Test 2

If you attempt to create two relationship fields with the same reverse slug, you should see, “A relationship field in this model has the same identifier.”

#### Test 3

If you attempt to create a relationship field that has a reverse slug of the same name as the slug of any field in the reference model, you should see, “A field in the X model has the same identifier.”

## Screenshots

Screenshot 2021-11-02 at 17 51 06

Screenshot 2021-11-02 at 17 51 31

## Documentation Changes

n/a

## Dependant PRs

n/a

## Notes
I think we may also need to confirm that reverse slugs do not use any of the reserved terms we added checks for in https://github.com/wpengine/atlas-content-modeler/pull/328, but that seems outside of the scope of this PR.

### DIFF
--- a/tests/acceptance/CreateContentModelRelationFieldCest.php
+++ b/tests/acceptance/CreateContentModelRelationFieldCest.php
@@ -1,4 +1,5 @@
 <?php
+use Codeception\Util\Locator;
 
 class CreateContentModelRelationFieldCest {
 
@@ -6,6 +7,13 @@ class CreateContentModelRelationFieldCest {
 		$i->maximizeWindow();
 		$i->loginAsAdmin();
 		$i->haveContentModel( 'Employee', 'Employees' );
+		$i->wait( 1 );
+		// Create a text field to check for reverse relationship conflicts with a reference model.
+		$i->click( 'Text', '.field-buttons' );
+		$i->fillField( [ 'name' => 'name' ], 'conflictsWithEmployeesSlug' );
+		$i->click( '.open-field button.primary' );
+		$i->wait( 1 );
+
 		$i->haveContentModel( 'Company', 'Companies' );
 		$i->wait( 1 );
 		$i->click( 'Relation', '.field-buttons' );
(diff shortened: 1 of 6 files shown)


ChrisWiegman (approved):

tests/acceptance/CreateContentModelRelationFieldCest.php
```diff
@@ -105,4 +113,52 @@ public function i_can_see_reverse_relationship_fields_when_clicking_the_appropri
...
+		$i->wait( 1 );
+		$i->see( 'A field in the employee model has the same identifier' );
+
+		// Resolve the reverse slug conflict and save the field.
+		$i->fillField( [ 'name' => 'reverseName' ], 'resolveConflict' );
+		$i->click( '.open-field button.primary' );
+		$i->wait( 1 );
+		$i->see( 'Company Employees', '.field-list div.widest' ); // Field was created successfully.
+
+		// Attempt to create a second relationship field with the same reverse
```
matthewguywright, line 134: Nice test!

matthewguywright (approved):

PR #346, opened 2021-11-09
[MTKA-1121] Add ability to set featured image using media field

ChrisWiegman:
## Description

Adds the ability to denote a media field as a "featured image"

Only one featured image can be selected per model and featured images are shown in REST, GraphQL and through other APIs in the same manner as they're done with standard content types like "post."

https://wpengine.atlassian.net/browse/MTKA-1121

## Testing

Appropriate testing has been included in both the phpunit and codeception suites.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 DOCKER_RUN       := docker run --rm
-COMPOSER_IMAGE   := -v "$$(pwd):/app" --user $$(id -u):$$(id -g) composer
+COMPOSER_IMAGE   := -v "$$(pwd):/app" --user $$(id -u):$$(id -g) composer:1
 NODE_IMAGE       := -w /home/node/app -v "$$(pwd):/home/node/app" --user node atlascontentmodeler_node_image
 HAS_CHROMEDRIVER := $(shell command -v chromedriver 2> /dev/null)
 CURRENTUSER      := $$(id -u)
--- a/includes/publisher/class-publisher-form-editing-experience.php
+++ b/includes/publisher/class-publisher-form-editing-experience.php
@@ -88,17 +88,29 @@ final class FormEditingExperience {
 		add_action( 'load-post.php', [ $this, 'feedback_notice_handler' ] );
 		add_action( 'load-post-new.php', [ $this, 'feedback_notice_handler' ] );
 		add_action( 'do_meta_boxes', [ $this, 'move_author_meta_box' ] );
+		add_action( 'do_meta_boxes', [ $this, 'remove_thumbnail_meta_box' ] );
 	}
 
 	/**
 	 * Removes unneeded post type features.
 	 */
 	public function remove_post_type_supports(): void {
-		foreach ( $this->models as $model => $info ) {
+		foreach ( $this->models as $model => $config ) {
 			remove_post_type_support( $model, 'editor' );
 			remove_post_type_support( $model, 'title' );
 			remove_post_type_support( $model, 'custom-fields' );
-			remove_post_type_support( $model, 'thumbnail' );
+			$remove_thumbnail = true;
+			if ( isset( $config['fields'] ) ) {
+				foreach ( $config['fields'] as $field ) {
+					if ( 'media' === $field['type'] && isset( $field['isFeatured'] ) && true === $field['isFeatured'] ) {
+						$remove_thumbnail = false;
+						continue;
+					}
+				}
+				if ( $remove_thumbnail ) {
+					remove_post_type_support( $model, 'thumbnail' );
+				}
+			}
 		}
 	}
 
@@ -367,12 +379,12 @@ final class FormEditingExperience {
 		);
 
 		foreach ( $all_field_slugs as $slug ) {
-			if ( ! array_key_exists( $slug, $posted_values ) ) {
-				$field_type = get_field_type_from_slug(
-					$slug,
-					$this->models[ $post->post_type ]['fields'] ?? []
-				);
+			$field_type = get_field_type_from_slug(
+				$slug,
+				$this->models[ $post->post_type ]['fields'] ?? []
+			);
 
+			if ( ! array_key_exists( $slug, $posted_values ) ) {
 				if ( 'relationship' === $field_type ) {
 					if ( ! in_array(
 						$slug,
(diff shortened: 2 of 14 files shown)


includes/publisher/class-publisher-form-editing-experience.php
```diff
@@ -88,17 +88,29 @@ public function bootstrap(): void {
...
 			remove_post_type_support( $model, 'editor' );
 			remove_post_type_support( $model, 'title' );
 			remove_post_type_support( $model, 'custom-fields' );
-			remove_post_type_support( $model, 'thumbnail' );
+			$remove_thumbnail = true;
+			if ( isset( $config['fields'] ) ) {
+				foreach ( $config['fields'] as $field ) {
+					if ( 'media' === $field['type'] && isset( $field['isFeatured'] ) && true === $field['isFeatured'] ) {
+						$remove_thumbnail = false;
+						continue;
```
mindctrl, line 107: My editor is pointing out that `continue` isn't necessary as the last statement in a loop. Maybe this should be `break`?

includes/publisher/class-publisher-form-editing-experience.php
```diff
@@ -610,6 +643,38 @@ public function filter_post_titles( string $title, int $id ) {
...
 
+	/**
+	 * Hides the “Screen Options” drop-down on post types registered by this plugin.
+	 *
+	 * @param bool       $show_screen The current state of the screen options dropdown.
+	 * @param \WP_Screen $screen Information about the current screen.
+	 *
+	 * @return bool The new state of the screen options dropdown. (False to disable.)
+	 */
+	public function hide_screen_options( bool $show_screen, $screen ): bool {
```
mindctrl, line 654: We removed this function in https://github.com/wpengine/atlas-content-modeler/commit/670c80ae1a62b051b989ad8b92cf9647186424bc

Probably left over from when you originally branched off `main`?

tests/acceptance/CreateContentModelMediaFieldCest.php
```diff
@@ -21,4 +21,89 @@ public function i_can_add_a_media_field_to_a_content_model( AcceptanceTester $i
 		$i->see( 'Media', '.field-list div.type' );
 		$i->see( 'Product Photo', '.field-list div.widest' );
 	}
+
+	/**
+	 * Ensure a user can add a featured image media field to the model and see it within the list.
+	 */
+	public function i_can_add_a_featured_imag_emedia_field_to_a_content_model( AcceptanceTester $i ) {
```
mindctrl, line 28:
```suggestion
	public function i_can_add_a_featured_image_media_field_to_a_content_model( AcceptanceTester $i ) {
```

mindctrl (changes requested):
Seems to be working well. A few minor things I see in the first pass.

On another note, I thought we had decided to limit the media picker to images only, like the description of the field says in DevEx - "Limits media selection to image types". But I don't see it called out specifically in the AC in Jira.

Makefile
```diff
@@ -1,5 +1,5 @@
 DOCKER_RUN       := docker run --rm
-COMPOSER_IMAGE   := -v "$$(pwd):/app" --user $$(id -u):$$(id -g) composer
+COMPOSER_IMAGE   := -v "$$(pwd):/app" --user $$(id -u):$$(id -g) composer:1
```
nickcernis, line 2: Thanks for fixing this here!

includes/publisher/lib/field-functions.php
```diff
@@ -74,7 +74,25 @@ function get_field_from_slug( string $slug, array $models, string $post_type ):
 }
 
 /**
- * Gets the field type from the field $slug.
+ * Determins if the field is the featured image or not.
```
nickcernis, line 77:
```suggestion
 * Determines if the field is the featured image or not.
```

includes/publisher/class-publisher-form-editing-experience.php
ChrisWiegman: Probably popped in while resolving merge conflicts. Fixed.

includes/publisher/class-publisher-form-editing-experience.php
ChrisWiegman: No notice in vscode but I've updated her anyway.

nickcernis (dismissed):
Working great. Confirmed adding, removing, changing fields, and the GraphQL and REST responses.

Thanks for your work on this, @ChrisWiegman!

Screenshot 2021-11-19 at 15 52 59

Screenshot 2021-11-19 at 15 51 16

includes/publisher/class-publisher-form-editing-experience.php
```diff
@@ -390,6 +402,27 @@ public function save_post( int $post_id, WP_Post $post ): void {
...
+				) &&
+				isset( $posted_values[ $slug ] )
+			) {
+				if ( '' === $posted_values[ $slug ] ) {
+					if ( ! delete_post_thumbnail( $post ) ) {
+						/* translators: %s: atlas content modeler field slug */
+						$this->error_save_post = sprintf( __( 'There was an error updating the %s field data.', 'atlas-content-modeler' ), $post->post_type );
+					}
+				} else {
+					delete_post_thumbnail( $post ); // Delete first is innefficient but avoids werid behavior of set which otherwise treats an existing thumbnail as a bug.
```
mindctrl, line 419: 'weird' is misspelled in the comment, but I wonder if the comment could be clearer? Is the bug that `set_post_thumbnail()` returns false if the value is the same as the one already stored in the db?

includes/publisher/class-publisher-form-editing-experience.php
```diff
@@ -390,6 +402,27 @@ public function save_post( int $post_id, WP_Post $post ): void {
...
+			if ( 'media' === $field_type &&
+				is_field_featured_image(
+					$slug,
+					$this->models[ $post->post_type ]['fields'] ?? []
+				) &&
+				isset( $posted_values[ $slug ] )
+			) {
+				if ( '' === $posted_values[ $slug ] ) {
+					if ( ! delete_post_thumbnail( $post ) ) {
+						/* translators: %s: atlas content modeler field slug */
```
mindctrl, line 415: Not sure it matters, but I think it's technically the model/post_type slug and not the field slug?

mindctrl (approved):

PR #348, opened 2021-11-10
[MTKA 1291] Enable saving relationships from the reverse side

nickcernis:
## Description

Allows related entries in a relationship field to be saved from the reverse side.

https://wpengine.atlassian.net/browse/MTKA-1291

## Testing

- Includes PHP unit tests for appending reverse relationship fields to the reference model.
- Extends existing e2e tests to check that entries can be saved from the reverse side, and that labels are correct.

### To test manually

1. Import the models below (you will need to remove the `.txt` extension).

[acm-models-export-2021-11-10-16-36-02.json.txt](https://github.com/wpengine/atlas-content-modeler/files/7514391/acm-models-export-2021-11-10-16-36-02.json.txt)

2. Create two “Left” posts named “Left 1” and “Left 2”, filling only the names.
3. Create a “Right” post named “Right 1” and link it to one or both of the left posts.

You should now see the Right post when you go to edit the Left posts.

You should be able to remove or edit the connections from either side.

## Screenshots

n/a

## Documentation Changes

n/a

## Dependant PRs

Targets #337 and should be merged into that first before it reaches main.

## Notes
Enforcing many-to-one and one-to-one from the reverse side will be tested and patched separately if required.

### DIFF
--- a/includes/publisher/class-publisher-form-editing-experience.php
+++ b/includes/publisher/class-publisher-form-editing-experience.php
@@ -344,13 +344,8 @@ final class FormEditingExperience {
 
 		// Sanitize field values.
 		foreach ( $posted_values as $field_id => &$field_value ) {
-			$field_id = sanitize_text_field( wp_unslash( $field_id ) ); // retains camelCase.
-
-			$field_type = get_field_type_from_slug(
-				$field_id,
-				$this->models[ $post->post_type ]['fields'] ?? []
-			);
-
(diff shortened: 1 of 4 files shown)


matthewguywright (approved):

mindctrl (approved):
Working well for me.

> Enforcing many-to-one and one-to-one from the reverse side will be tested and patched separately if required.

I think we need to do this for one-to-many and many-to-one to reverse them on the other side. Otherwise it appears the UI doesn't seem to enforce the proper cardinality.

PR #354, opened 2021-11-11
Add PHPUnit Polyfills library

mindctrl:
## Description
This adds the PHPUnit Polyfills library to the Composer dev dependencies, which fixes an issue with tests failing as a result of changes to the WordPress PHPUnit test suite.

Also renames the `setUp` and `tearDown` fixture methods to `set_up` and `tear_down`, per the instructions in the make.wp post linked below.

See:
https://make.wordpress.org/core/2021/09/27/changes-to-the-wordpress-core-php-test-suite/

https://github.com/wpengine/atlas-content-modeler/pull/337#issuecomment-966094383

### DIFF
--- a/tests/integration/updater/test-version-updater.php
+++ b/tests/integration/updater/test-version-updater.php
@@ -14,8 +14,8 @@ class UpdaterTestCases extends WP_UnitTestCase {
 
 	protected $versions;
 
-	public function setUp(): void {
-		parent::setUp();
+	public function set_up(): void {
+		parent::set_up();
 
 		$file_data      = get_file_data( ATLAS_CONTENT_MODELER_FILE, array( 'Version' => 'Version' ) );
 		$plugin_version = $file_data['Version'];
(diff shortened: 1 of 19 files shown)


tests/integration/updater/test-version-updater.php
```diff
@@ -47,6 +47,7 @@ public function test_update_plugin_installed_new_version(): void {
 	 * @covers ::\WPE\AtlasContentModeler\VersionUpdater\update_plugin()
 	 */
 	public function test_update_plugin_no_saved_version(): void {
+		delete_option( 'atlas_content_modeler_current_version' );
```
mindctrl, line 50: I added this because simply renaming `tearDown` to `tear_down` caused this test to start failing, at least in my local environment, despite the `tear_down` method running as expected before this test runs. I tried called `parent::tear_down()` after the `delete_option` call in the renamed `tear_down` method, but that didn't fix it. Am I missing something?

ChrisWiegman: I don't see that you're missing anything but in quick testing I can reproduce the problem. I'm wondering it it is a timing thing/race condition given the I/O latency to the DB?

ChrisWiegman (approved):
Tests are passing so I'm inclined to think it worth merging this.

PR #355, opened 2021-11-11
MTKA-1116: Add the ability to edit entry slugs

matthewguywright:
## Description
Enable slug editing for publisher.

## Testing
1. Create model
2. Go to publisher
3. Make sure screen options is displaying
4. Open screen options and enable "slug"
5. You should see the slug metabox in the sidebar

## Screenshots
Screen Shot 2021-11-11 at 11 05 16 AM

### DIFF
--- a/includes/publisher/class-publisher-form-editing-experience.php
+++ b/includes/publisher/class-publisher-form-editing-experience.php
@@ -613,22 +613,6 @@ final class FormEditingExperience {
 	}
 
 	/**
-	 * Hides the “Screen Options” drop-down on post types registered by this plugin.
-	 *
-	 * @param bool       $show_screen The current state of the screen options dropdown.
-	 * @param \WP_Screen $screen Information about the current screen.
-	 *
-	 * @return bool The new state of the screen options dropdown. (False to disable.)
-	 */
-	public function hide_screen_options( bool $show_screen, $screen ): bool {
-		if ( in_array( $screen->post_type, array_keys( $this->models ), true ) ) {
-			return false;
-		}
-
-		return $show_screen;
-	}
-
-	/**
 	 * Moves the meta boxes to the sidebar.
 	 *
 	 * Improves usability by moving the meta box away from the main editor area
--- a/includes/publisher/class-publisher-form-editing-experience.php
+++ b/includes/publisher/class-publisher-form-editing-experience.php
@@ -84,10 +84,9 @@ final class FormEditingExperience {
 		add_filter( 'redirect_post_location', [ $this, 'append_error_to_location' ], 10, 2 );
 		add_action( 'admin_notices', [ $this, 'display_save_post_errors' ] );
 		add_filter( 'the_title', [ $this, 'filter_post_titles' ], 10, 2 );
-		add_filter( 'screen_options_show_screen', [ $this, 'hide_screen_options' ], 10, 2 );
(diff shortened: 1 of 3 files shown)


tests/acceptance/PublishModelCest.php
```diff
@@ -210,8 +210,13 @@ public function i_can_publish_a_model_entry( AcceptanceTester $i ) {
 		$i->click( 'Add New', '.wrap' );
 		$i->wait( 1 );
 
-		// Check that the author meta box was moved to the sidebar by default.
+		$i->click( 'Screen Options', '#screen-options-link-wrap' );
+		$i->wait( 1 );
+		$i->checkOption( 'slugdiv-hide' );
```
mindctrl, line 215: @matthewguywright  The `checkOption` click failed because the feedback banner was covering it up...

> element click intercepted: Element <button type="button" id="show-settings-link" class="button show-settings" aria-controls="screen-options-wrap" aria-expanded="false">...</button> is not clickable at point (695, 47). Other element would receive the click: <div id="feedbackBanner" style="background-color: #F2EFFD;" class="atlas-content-modeler notice notice-info is-dismissible">...</div>

It might help to adjust the screen size here. Maybe something like `$i->resizeWindow( 1024, 1024 );` in the `_before` method above the `maximizeWindow` call, like we've done in some other tests.

nickcernis (commented):
@matthewguywright Working well for me. 👍 

Might it be worth adjusting the top margin of the feedback banner so it doesn't overlay the screen options tab we reintroduced in this PR?

If we want to direct people to the Screen Options tab in docs/support it might be easier if they can read it:

Screenshot 2021-11-15 at 13 01 03

includes/publisher/class-publisher-form-editing-experience.php
```diff
@@ -84,10 +84,9 @@ public function bootstrap(): void {
 		add_filter( 'redirect_post_location', [ $this, 'append_error_to_location' ], 10, 2 );
 		add_action( 'admin_notices', [ $this, 'display_save_post_errors' ] );
 		add_filter( 'the_title', [ $this, 'filter_post_titles' ], 10, 2 );
-		add_filter( 'screen_options_show_screen', [ $this, 'hide_screen_options' ], 10, 2 );
```
nickcernis, line 87: Looks like we can also remove the `hide_screen_options` function from the same file. It's not used anywhere else that I can see.

mindctrl:
> Might it be worth adjusting the top margin of the feedback banner so it doesn't overlay the screen options tab we reintroduced in this PR?

Ah yes, if it's affecting us in normal use, I think we should adjust it here.

tests/acceptance/PublishModelCest.php
```diff
@@ -210,8 +211,16 @@ public function i_can_publish_a_model_entry( AcceptanceTester $i ) {
 		$i->click( 'Add New', '.wrap' );
 		$i->wait( 1 );
 
-		// Check that the author meta box was moved to the sidebar by default.
+		$i->click( '.notice-dismiss', '#feedbackBanner' );
+		$i->wait( 1 );
```
mindctrl, line 215:
```suggestion
```

@matthewguywright if you remove this addition from the test, it'll catch any style regressions that cause the banner to overlay the Screen Options tab.

mindctrl (approved):

PR #361, opened 2021-11-24
[NOJIRA] Add data about duplicate multiple choice option slugs to the error directly

nickcernis:
FAO @anthonyburchell.

An example of adding data about the duplicate option slugs to the returned validation error data directly.

This removes the need to call `$error->add()`, which addresses @mindctrl's comment at https://github.com/wpengine/atlas-content-modeler/pull/210#discussion_r752319640.

A similar pattern could be used to replace the other `add()` calls in `includes/rest-api/routes/content-model-field.php`.


anthonyburchell (approved):
Thanks so much, Nick!

PR #364, opened 2021-11-29
[MTKA-1105] Remove dead code

nickcernis:
## Description

- Removes dead code identified by @apmatthews during an earlier code review. (Thanks, Andrew!)
- Removes unused PHP/JS imports.

Note that these suggestions from the Jira ticket no longer apply to the current codebase:

- Remove the clickHandler function in includes/publisher/js/src/App.jsx (no longer exists).
- Unused “use” statement for “generate_custom_post_type_args” in rest-api-endpoint-registration.php (no longer exists).
- The `delete_model()` function in rest-api-endpoint-registration.php is never used. (`delete_model()` was moved to `includes/rest-api/models.php` and is now imported and used in `includes/rest-api/routes/content-model.php`).
- The enqueued script and should_show_feedback_banner() function in settings-callbacks.php is unused. The feedback banner script is enqueued on the developer side, but the banner is not used there. There's a permanent feedback button instead, which I assume takes the place of the banner.  (`should_show_feedback_banner()` is now used in that file, and the feedback button script is used on the publisher and developer side — on the developer side the feedback button is removed the first time it is clicked.)

These suggestions have been applied:

- [x] FormEditingExperience is bootstrapped twice.
    - [x] The constructor calls the bootstrapping function, but we also call it manually in atlas-content-modler.php after instantiating the class. We should decide which approach makes more sense and remove the other. 
- [x] Duplicate input fields in publisher/js/src/components/MediaUploader.jsx (currently on lines 71 & 123)
    - [x] Both are hidden text fields with name=atlas-content-modeler[${modelSlug}][${field.slug}] 
- [x] Remove the entire includes/publisher/js/src/ActionButtons.jsx file 

https://wpengine.atlassian.net/browse/MTKA-1105

## Testing

Existing tests should pass as before.

## Screenshots

n/a

## Documentation Changes

n/a
 
## Dependant PRs

n/a


mindctrl (dismissed):
LGTM.

Noting that the double bootstrap could also be removed from the test that was just adjusted here: https://github.com/wpengine/atlas-content-modeler/pull/362/files#diff-deeee92b1cd711e2522a44a759065175c27362a1ac06beacccd655e21b2c5f72R29

nickcernis:
> LGTM.
> 
> Noting that the double bootstrap could also be removed from the test that was just adjusted here: https://github.com/wpengine/atlas-content-modeler/pull/362/files#diff-deeee92b1cd711e2522a44a759065175c27362a1ac06beacccd655e21b2c5f72R29

Good call, thanks @mindctrl. Adjusted in https://github.com/wpengine/atlas-content-modeler/pull/364/commits/ef562a73d78278d4f7fa8ff411d858c3857173d8.

mindctrl (approved):

PR #365, opened 2021-11-30
[MTKA 1089] Standardize error codes

nickcernis:
## Description

Uses the `acm_` prefix for all error codes, instead of a mix of `wpe_` and `atlas_content_modeler_`

https://wpengine.atlassian.net/browse/MTKA-1089

## Testing

Existing tests should pass.

## Screenshots

n/a

## Documentation Changes

It's unlikely that people are using these APIs directly, but we should mention this in our changelog even though it's effectively an internal change.

## Dependant PRs

n/a

### DIFF
--- a/includes/rest-api/models.php
+++ b/includes/rest-api/models.php
@@ -34,7 +34,7 @@ function create_model( string $post_type_slug, array $args ) {
 
 	if ( ! empty( $content_types[ $args['slug'] ] ) || array_key_exists( $args['slug'], $existing_content_types ) ) {
 		return new WP_Error(
-			'atlas_content_modeler_already_exists',
+			'acm_already_exists',
 			__( 'A content model with this Model ID already exists.', 'atlas-content-modeler' ),
 			[ 'status' => 400 ]
 		);
@@ -76,7 +76,7 @@ function create_models( array $models ) {
 
 		if ( ! empty( $content_types[ $args['slug'] ] ) || array_key_exists( $args['slug'], $existing_content_types ) ) {
 			return new WP_Error(
-				'atlas_content_modeler_already_exists',
+				'acm_already_exists',
 				// translators: The name of the model.
 				sprintf( __( 'A model with slug ‘%s’ already exists.', 'atlas-content-modeler' ), $args['slug'] ),
 				[ 'status' => 400 ]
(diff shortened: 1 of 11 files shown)


includes/rest-api/models.php
```diff
@@ -34,7 +34,7 @@ function create_model( string $post_type_slug, array $args ) {
 
 	if ( ! empty( $content_types[ $args['slug'] ] ) || array_key_exists( $args['slug'], $existing_content_types ) ) {
 		return new WP_Error(
-			'atlas_content_modeler_already_exists',
+			'acm_already_exists',
```
mindctrl, line 37: The shorter version made it more obvious to me that "already_exists" isn't a great error code. What already exists? I know by reading the surrounding code, but it might be nice for the code to be a bit more descriptive.

includes/rest-api/models.php
```diff
@@ -108,15 +108,15 @@ function get_model_args( string $post_type_slug, array $args ) {
...
+			'acm_invalid_id',
 			__( 'Please provide a valid Model ID.', 'atlas-content-modeler' ),
 			[ 'status' => 400 ]
 		);
 	}
 
 	if ( empty( $args['singular'] ) || empty( $args['plural'] ) ) {
 		return new WP_Error(
-			'atlas_content_modeler_invalid_labels',
+			'acm_invalid_labels',
```
mindctrl, line 119: Pre-existing, but we use a different error code here than we do in `update_model()` below (`acm_invalid_content_model_arguments`) when the labels are missing. We send back the same text, but a different error code. Thoughts on making them the same code?

includes/rest-api/models.php
```diff
@@ -202,7 +202,7 @@ function delete_model( string $post_type_slug ) {
 
 	if ( empty( $post_type_slug ) || empty( $content_types[ $post_type_slug ] ) ) {
 		return new WP_Error(
-			'atlas_content_modeler_invalid_id',
+			'acm_invalid_id',
```
mindctrl, line 205: Thoughts on making this `acm_invalid_model_id`?

mindctrl (dismissed):
Looks good, but left a few questions for us to consider.

includes/rest-api/models.php
nickcernis: Good idea. I changed it to `acm_model_exists` in 76261e4.

includes/rest-api/models.php
nickcernis: Tweaked in 6ce9076. It meant also adding `acm_invalid_taxonomy_id` for the taxonomy endpoint.

includes/rest-api/models.php
nickcernis: Nice idea, I went with `acm_invalid_labels` in e2959fa, which is consistent with the error code from the taxonomy endpoint.

mindctrl (approved):

PR #377, opened 2021-12-07
MTKA-1333: replace deprecated CircleCI Docker images

mindctrl:
## Description
Replaces the legacy CircleCI Docker images with the latest supported versions.

CircleCI is deprecating some existing Docker images and replaced them with newer, faster images. Legacy convenience images with the prefix `circleci/` will be deprecated on Dec. 31, 2021.

https://wpengine.atlassian.net/browse/MTKA-1333

## Testing
Verify that CircleCI tests still run and pass as expected.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,8 +5,8 @@
 version: 2.1
 orbs:
   wp-product-orb: wpengine/wp-product-orb@1.3.0
-  php: cimg/php@7.4.26
-  node: cimg/node@16.13.1
+  php: cimg/php:7.4.26
+  node: cimg/node:16.13.1
 
 commands:
   set_version_variable:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -228,15 +229,12 @@ jobs:
       - attach_workspace:
           at: .
 
+      - browser-tools/install-chrome
```
mindctrl, line 232: Without this, Chrome was no longer available. See: https://circleci.com/developer/images/image/cimg/php#browsers

mindctrl: This could also be used to install `chrome-driver`, allowing us to remove our custom job, but I think that should be done in a different PR.

nickcernis (approved):

PR #382, opened 2021-12-10
Poc/repeatable lists use state

anthonyburchell:
## Description


anthonyburchell:
Amazing stuff. Thanks Nick! That simplifies it greatly :D

PR #387, opened 2021-12-15
MTKA-1332: model/taxonomy label check for collisions

matthewguywright:
[Edited by Nick on 11 Jan 2021]

## Description

Ensures that models and taxonomies can not be created or updated if:

- An ACM singular or plural model label collides with another model or taxonomy label. (GraphQL registers them at the top level, so labels must be globally unique.)
- The singular or plural label of the model or taxonomy would collide with a reserved top-level GraphQL “field”, such as 'post', 'posts', 'category', categories', 'plugin', 'plugins' etc. 

Screenshot 2022-01-11 at 14 10 19

Screenshot 2022-01-11 at 14 11 19

Also checks models that are bulk imported via the Import Tool. If any model in the group contains a bad label the full import will be rejected so that the user can correct the file and re-import. (I chose to reject the import in full so the user doesn't have to trim any models from their import file that were correctly imported the first time.)

Screenshot 2022-01-11 at 13 58 42

https://wpengine.atlassian.net/browse/MTKA-1332

## Automated Tests
Includes:
- Unit tests for the REST endpoint updates.
- End-to-end tests to confirm user feedback is provided if an illegal label is used.

## Manual Testing
You should no longer be able to create a model or taxonomy with the same single or plural label as another model or taxonomy. (This requires that WPGraphQL is active.)

You should not be able to create models or taxonomies with a singular or plural label that uses a reserved term such as "post", "posts", "plugin" or "plugins". (This part should work even if WPGraphQL is inactive.)

### DIFF
--- a/tests/integration/api-validation/test-rest-models-endpoints.php
+++ b/tests/integration/api-validation/test-rest-models-endpoints.php
@@ -18,8 +18,8 @@ class RestModelsEndpointTests extends WP_UnitTestCase {
 	private $route     = '/atlas/content-models';
 	private $test_models;
 
-	public function set_up(): void {
-		parent::set_up();
+	public function setUp(): void {
+		parent::setUp();
 
 		$this->test_models = $this->get_models();
 
@@ -43,8 +43,8 @@ class RestModelsEndpointTests extends WP_UnitTestCase {
 		do_action( 'rest_api_init' );
 	}
 
-	public function tear_down() {
-		parent::tear_down();
+	public function tearDown() {
+		parent::tearDown();
 		wp_set_current_user( null );
 		global $wp_rest_server;
 		$wp_rest_server = null;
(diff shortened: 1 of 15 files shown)


tests/integration/api-validation/test-rest-models-endpoints.php
```diff
@@ -18,8 +18,8 @@ class RestModelsEndpointTests extends WP_UnitTestCase {
 	private $route     = '/atlas/content-models';
 	private $test_models;
 
-	public function set_up(): void {
-		parent::set_up();
+	public function setUp(): void {
+		parent::setUp();
```
mindctrl, line 22: Did you run into an issue that required this to be renamed? I remember something quirky with these methods when we renamed them [here](https://github.com/wpengine/atlas-content-modeler/pull/354), but I can't recall specifics.

nickcernis: @mindctrl Ah, sorry, I didn't mean to commit this change. I can't run these tests locally without changing it back to `setUp` and `tearDown` but I haven't been able to figure out why yet. (I think @anthonyburchell had the same issue locally.)

I'll change it back for consistency and work around it locally for now.

nickcernis: Changed in https://github.com/wpengine/atlas-content-modeler/pull/387/commits/7b99b80384f7274e1b38dc773b0202df7d524848.

mindctrl: > I can't run these tests locally without changing it back to setUp and tearDown but I haven't been able to figure out why yet. 

Oh yeah, I think I saw the same previously. Thanks for the reminder.

mindctrl (approved):

tests/integration/api-validation/test-rest-models-endpoints.php
nickcernis: @mindctrl Interestingly, if I make the change recommended in [option 1 of the PHPUnit-Polyfills docs](https://github.com/Yoast/PHPUnit-Polyfills#testcases) all runs well locally:

```diff
diff --git a/tests/integration/api-validation/test-rest-models-endpoints.php b/tests/integration/api-validation/test-rest-models-endpoints.php
index 0a59a62..0c7247a 100644
--- a/tests/integration/api-validation/test-rest-models-endpoints.php
+++ b/tests/integration/api-validation/test-rest-models-endpoints.php
@@ -6,7 +6,7 @@ use function WPE\AtlasContentModeler\ContentRegistration\get_registered_content_
 
 require_once __DIR__ . '/test-data/fields.php';
 
-class RestModelsEndpointTests extends WP_UnitTestCase {
+class RestModelsEndpointTests extends Yoast\PHPUnitPolyfills\TestCases\TestCase {
 
 	/**
 	 * The REST API server instance.
```

So it seems that the polyfill lib isn't automatically extending `WP_UnitTestCase` with the missing `set_up` and `tear_down` method in some cases.

Intelephense also shows `set_up` as an undefined method when extending `WP_UnitTestCase`:

Screenshot 2022-01-12 at 16 01 28

Clicking `WP_UnitTestCase` takes me to `wp-content/plugins/atlas-content-modeler/vendor/lucatume/wp-browser/src/includes/testcase.php`, so my guess is it's some kind of conflict with that library.

The blog post you linked to above says:

> You can still extend the WP_UnitTestCase class for your tests and will receive access to everything which was available before + more (i.e. a complete set of polyfills).

But that doesn't seem to be the case consistently. It's also not clear if that was supposed to be the case for WP core dev or plugin dev or both.

Issue #386, opened 2021-12-15
Publishing content in a new Model leads to 404s


jasonbahl:
A clear and concise description of the bug goes here.

### To reproduce
1. Create a new public model using ACM
2. Create a new entry in that Model
3. Visit the post of that model
4. See 404

### Expected behavior
I expect to be able to create a model in ACM, then publish content of that model and have it publicly visible. 

Currently, I have to go to manually flush permalinks after creating a model. 

### Possible solution
When creating / updating / deleting ACM models, permalinks should automatically be flushed by ACM.

### Screenshots
I recorded a video walking through some things I was experiencing with ACM and this issue is covered in the video.
https://www.youtube.com/watch?v=SmmGqy7bXxo

### Version information
- Plugin version: 0.11.0
- WordPress version: 5.8.2
- Operating system: Mac OSX 11.6.1
- Browser: Chrome v96.0.4664.93

mindctrl:
Nice catch. Thanks for reporting this.

nickcernis:
Also logged at https://wpengine.atlassian.net/browse/MTKA-1364.

nickcernis:
Fixed by https://github.com/wpengine/atlas-content-modeler/pull/390 (unreleased, but scheduled for 0.13.0). Thanks again for the report!

PR #390, opened 2022-01-05
MTKA-1364: Flush rewrite rules when adding/removing models

rfmeier:
## Description

If a Content Model was `public` and entries were created, they could not be viewed on the front-end due to rewrite rules not being flushed. This pull request adds [flush_rewrite_rules()](https://developer.wordpress.org/reference/functions/flush_rewrite_rules/) when a Content Model is created, deleted, or updated.

### DIFF
--- a/includes/content-registration/custom-post-types-registration.php
+++ b/includes/content-registration/custom-post-types-registration.php
@@ -403,7 +403,13 @@ function get_registered_content_types(): array {
  * @return bool
  */
 function update_registered_content_types( array $args ): bool {
-	return update_option( 'atlas_content_modeler_post_types', $args );
+	$updated = update_option( 'atlas_content_modeler_post_types', $args );
+
+	if ( $updated ) {
+		flush_rewrite_rules();
+	}
+
+	return $updated;
 }
 
 /**
--- a/includes/content-registration/custom-post-types-registration.php
+++ b/includes/content-registration/custom-post-types-registration.php
@@ -406,7 +406,7 @@ function update_registered_content_types( array $args ): bool {
 	$updated = update_option( 'atlas_content_modeler_post_types', $args );
 
 	if ( $updated ) {
-		flush_rewrite_rules();
+		flush_rewrite_rules( false );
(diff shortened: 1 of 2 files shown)


nickcernis (dismissed):
Working well for me.

- [x] Confirmed public model entries no longer generate a 404. 
- [x] Did not notice any measurable delay introduced in flushing rewrite rules for every change of the `update_registered_content_types` option. (I wondered if this would slow field creation and other operations that touch `atlas_content_modeler_post_types` but don't need a rewrite rule flush, but it seems fine.)

Noting that this fixes https://github.com/wpengine/atlas-content-modeler/issues/386.

includes/content-registration/custom-post-types-registration.php
```diff
@@ -403,7 +403,13 @@ function get_registered_content_types(): array {
  * @return bool
  */
 function update_registered_content_types( array $args ): bool {
-	return update_option( 'atlas_content_modeler_post_types', $args );
+	$updated = update_option( 'atlas_content_modeler_post_types', $args );
+
+	if ( $updated ) {
+		flush_rewrite_rules();
```
mindctrl, line 409: Is it necessary to let it rewrite the `.htaccess`? `flush_rewrite_rules( false )` would prevent that.

mindctrl:
> flushing rewrite rules for every change

We could put it inside the REST callback functions instead, but I think due to timing it would require some additional tweaks like running running `register_post_type` to register the new post type and `flush_rewrite_rules()` on the `shutdown` hook or something like that. All that said, this seems ok as is. We can assess and iterate if necessary.

includes/content-registration/custom-post-types-registration.php
rfmeier: This is a good catch. My guess is a hard rewrite shouldn't be needed. I don't believe post types are written to the `.htaccess` file. It's been years since I have used Apache, but I will check it out.

mindctrl (approved):

PR #397, opened 2022-01-10
MTKA-1326: Add Emotion npm package

rfmeier:
## Description

This pull request adds the [Emotion](https://emotion.sh/docs/introduction) npm package for adding inline styling to React components.

Note: Currently works with using [JSX Pragma](https://emotion.sh/docs/css-prop#jsx-pragma). The babel integration was causing issues and not working. This could be due to WordPress's React version `<= 17` and could be reviewed again in the future.


nickcernis (approved):
Looks good! I couldn't get it to work without the pragma either.

PR #403, opened 2022-01-18
Adds single repeatable text fields [SETTINGS]

anthonyburchell:
This PR adds single repeatable text fields using the `isRepeatable` property set in the content modeler field settings for text.


includes/publisher/lib/field-functions.php
```diff
@@ -177,12 +192,16 @@ function append_reverse_relationship_fields( array $models, string $post_type ):
  *
  * @param string $type The type of field.
  * @param mixed  $value The unsanitized field value already processed by `wp_unslash()`.
+ * @param bool   $field_repeatable If the current field is repeatable.
  *
  * @return mixed The sanitized field value.
  */
-function sanitize_field( string $type, $value ) {
+function sanitize_field( string $type, $value, $field_repeatable ) {
```
mindctrl, line 199: It might be helpful if this function accepted the `$field` array instead of only the repeatable value, as there will likely be other use cases for knowing the field configuration.

includes/publisher/lib/field-functions.php
```diff
@@ -106,6 +106,21 @@ function get_field_type_from_slug( string $slug, array $models, string $post_typ
...
 }
 
+/**
+ * Gets the field property of isRepeatable from the field slug.
+ *
+ * @param string $slug Field slug to look for the 'type' property.
+ * @param array  $models Models with field properties to search for the `$slug`.
+ * @param string $post_type Current post type on the publisher screen.
+ *
+ * @return string Field type if found, or 'unknown'.
```
mindctrl, line 116:
```suggestion
 * @return bool Whether or not the field is repeatable.
```

includes/publisher/lib/field-functions.php
```diff
@@ -106,6 +106,21 @@ function get_field_type_from_slug( string $slug, array $models, string $post_typ
...
+ */
+function get_field_repeatable_from_slug( string $slug, array $models, string $post_type ): bool {
```
mindctrl, line 118: Maybe this function's purpose would be clearer with a name like `is_field_repeatable()` or something like that?

includes/settings/js/src/components/fields/TextFields.jsx
```diff
@@ -111,13 +137,20 @@ const TextFields = ({ register, data, editing, fieldId }) => {
 							</span>
 						</label>
 					</div>
-					<div className="radio-row">
+					<div
+						className={
+							!showRepeatable
+								? "d-flex flex-column d-sm-flex flex-sm-row"
```
mindctrl, line 143: Removing the `radio-row` class causes a display issue. Screenshot:
image

includes/publisher/lib/field-functions.php
nickcernis: We might not need to pass an extra argument here.

We could check if `$value` is an array (signalling that it's repeatable) or a string (signalling that it's not). That approach seemed to work in my POC:

https://github.com/wpengine/atlas-content-modeler/blob/79d3bf3be731d22fa9a9c3237b4c84853e7a9846/includes/publisher/lib/field-functions.php#L183-L190

mindctrl: Nice, I like that approach over adjusting the function signature at this point.

mindctrl:
I think checking the "Make this field repeatable" box should hide/disable the Title Field checkbox, as I don't think it will make sense to have a repeating title field.

includes/content-registration/custom-post-types-registration.php
```diff
@@ -495,6 +495,10 @@ function register_content_fields_with_graphql( TypeRegistry $type_registry ) {
 				$field['type'] = array( 'list_of' => 'String' );
 			}
 
+			if ( $field['isRepeatable'] ) {
+				$field['type'] = array( 'list_of' => 'String' );
```
mindctrl, line 499: There's a pattern emerging here, where the `map_html_field_type_to_graphql_field_type()` function can't set the `$field['type']` value by itself because it needs additional context from the field configuration, which results in us overriding the field type based on the two separate conditions above (one for list, one for repeatable). Just noting for future reference that we need to make some adjustments so the field type can be set without the extra logic outside the mapping function. I'll add a separate issue for this.

includes/publisher/lib/field-functions.php
anthonyburchell: good call. in my testing this works much better. :) thanks, Nick!

includes/publisher/lib/field-functions.php
anthonyburchell: yeah agreed. since we're checking array though this is no longer needed.

mindctrl:
@anthonyburchell can you bring `feat/text-repeater-field` up to date with `main`? That should clean up the diff view and make it easier to see only the requested changes.

includes/publisher/lib/field-functions.php
```diff
@@ -183,6 +183,9 @@ function append_reverse_relationship_fields( array $models, string $post_type ):
 function sanitize_field( string $type, $value ) {
 	switch ( $type ) {
 		case 'text':
+			if ( is_array( $value ) ) {
+				return $value;
```
nickcernis, line 187: Probably worth sanitizing the values here so they're treated the same way as the single value:

```suggestion
				return array_map( 'wp_strip_all_tags', $value );
```

includes/settings/js/src/components/fields/TextFields.jsx
```diff
@@ -14,19 +14,48 @@ const TextFields = ({ register, data, editing, fieldId }) => {
...
-
+	const [showRepeatable, setShowRepeatable] = useState(
+		data?.isRepeatable === true
+	);
 	return (
 		<>
-			{showTitleField && (
+			{data && (
 				<div className="field">
+					<legend>Repeatable Field</legend>
```
nickcernis, line 24: Needs to be translatable with `__()`.

includes/settings/js/src/components/fields/TextFields.jsx
```diff
@@ -14,19 +14,48 @@ const TextFields = ({ register, data, editing, fieldId }) => {
...
-
+	const [showRepeatable, setShowRepeatable] = useState(
+		data?.isRepeatable === true
+	);
 	return (
 		<>
-			{showTitleField && (
+			{data && (
 				<div className="field">
+					<legend>{__("Repeatable Field")}</legend>
```
nickcernis, line 24: @anthonyburchell These need the "domain" too: 
```suggestion
					<legend>
						{__("Repeatable Field", "atlas-content-modeler")}
					</legend>
```

mindctrl (approved):
LGTM. Once the text domain thing @nickcernis mentioned is handled, I think we can merge this into the feature branch and focus on the publisher side.

nickcernis (approved):

PR #413, opened 2022-01-26
Add Repeatable Text Field [Publisher]

anthonyburchell:
This PR adds the publisher side changes for repeatable text fields. ✨


mindctrl:
The min/max settings don't seem to be enforced on repeating fields.

mindctrl:
The "required" field setting doesn't seem to be enforced.

mindctrl:
Bringing over from Slack so we don't lose it:

Just noticed it's possible to set a repeating title field, which causes WP to throw some PHP notices.
To reproduce:
* Add text field to model
* Check "Use this field as the entry title"
* Check "Make this field repeatable"

PHP errors happen because WP expects the title to be a string, not array.

Checking "entry title" should prevent "repeatable" from being checked. It works the other way around, just need some additional logic to make it work both ways.

includes/publisher/scss/_text-field.scss
```diff
@@ -0,0 +1,133 @@
...
+		.repeater-input input {
+			width: 100% !important;
+			margin: 0px !important;
+		}
+	}
+
+	.repeater-input-container input[type="text"] {
+		margin: 0px !important;
+		width: 100% !important;
+	}
```
mindctrl, line 35: Are all the `!important`s necessary in this file?

includes/publisher/js/src/components/Text.jsx
```diff
@@ -0,0 +1,276 @@
...
+														<div
+															className="me-lg-1 repeater-input-container flex-fill"
+															name="repeaters"
+														>
+															{field.inputType ===
+															"single" ? (
+																<input
+																	name={`atlas-content-modeler[${modelSlug}][${field.slug}][${index}]`}
+																	placeholder={__(
+																		`Add ${field.name}`,
```
mindctrl, line 56: An interesting side effect of using the field name is this:
image

Basically it assumes the label will always make a readable sentence/fragment. In the case of the label I used, it resulted in "add" being used twice in a row. I think we should consider changing this string to something like "Add another", so that it's generic.

nickcernis: “Add another” works for the button.

For the input placeholder text, perhaps we remove the initial “Add” so it just uses the label?

includes/publisher/js/src/components/Field.jsx
```diff
@@ -97,238 +98,7 @@ function fieldMarkup(field, modelSlug, errors, validate) {
...
-					)}
-					<span className="error">
-						<Icon type="error" />
-						<span role="alert">
-							{errors[field.slug] ?? defaultError}
-						</span>
-					</span>
-				</>
-			);
+			return <Text field={field} modelSlug={modelSlug} />;
```
nickcernis, line 101: We need to pass in `errors={errors}`, `validate={validate}` and `defaultError={defaultError}` here.

includes/publisher/js/src/components/Text.jsx
```diff
@@ -0,0 +1,280 @@
+/* global atlasContentModelerFormEditingExperience */
+import React, { useState, useRef } from "react";
+import Icon from "../../../../components/icons";
+import { __ } from "@wordpress/i18n";
+const { wp } = window;
+import AddIcon from "../../../../components/icons/AddIcon";
+import TrashIcon from "../../../../components/icons/TrashIcon";
+
+export default function Text({ field, modelSlug, defaultError }) {
+	const [errors, setErrors] = useState({});
```
nickcernis, line 10: This needs to receive the extra props.

This also lets us remove the `errors` useState, because errors is passed in from the parent component.

```suggestion
export default function Text({
	field,
	errors,
	validate,
	modelSlug,
	defaultError,
}) {
```

nickcernis (commented):
This is looking so good!

I left a couple of small code change suggestions above.

Bringing this branch up to date with main should solve the failing `plugin-test-content-connect` job.

There are some failing tests for `plugin-test-e2e` and `plugin-test-jest` but worth checking if they're still there after updating from main before digging in deeper.

includes/publisher/lib/field-functions.php
```diff
@@ -184,7 +184,13 @@ function sanitize_field( string $type, $value ) {
...
 		case 'text':
 			if ( is_array( $value ) ) {
-				return array_map( 'wp_strip_all_tags', $value );
+				$repeater_text_value = [];
+				foreach ( $value as $text_string ) {
+					if ( $text_string !== '' ) {
+						$repeater_text_value[] = $text_string;
+					}
+				}
+				return array_map( 'wp_strip_all_tags', $repeater_text_value );
```
mindctrl, line 193:
```suggestion
				return array_filter( array_map( 'wp_strip_all_tags', $value ) );
```

We should be able to remove empty items with `array_filter`.

includes/publisher/scss/_text-field.scss
```diff
@@ -0,0 +1,133 @@
+@import "../../../includes/shared-assets/scss/_variables";
+
+.text {
+	.text-table {
+		// display: inline-flex;
```
mindctrl, line 5:
```suggestion
```

nickcernis (approved):

PR #419, opened 2022-02-02
MTKA-1383: emotionize dropdown

matthewguywright:
## Description
Update Dropdowns on settings and publisher side to utilize Emotion styling.


includes/settings/js/src/components/TaxonomiesDropdown.jsx
```diff
@@ -63,7 +64,7 @@ export const TaxonomiesDropdown = ({ taxonomy }) => {
 	}, [timer]);
 
 	return (
-		<span className="dropdown">
+		<Dropdown className="dropdown">
```
rfmeier, line 67: Is the `dropdown` class still needed?

includes/settings/scss/_model-list.scss
```diff
@@ -125,6 +75,14 @@
...
 }
 
+@media (max-width: $breakpoint-tablet) {
+	.neg-margin-wrapper {
+		.dropdown {
+			margin-left: -80px !important;
+		}
+	}
+}
+
```
nickcernis, line 85: Is there a way we can move the media query to the component/Emotion styles?

includes/settings/js/src/components/TaxonomiesDropdown.jsx
matthewguywright: Updated. It was used for the media query that was outside this emotion component. That has been moved internal.

includes/settings/scss/_model-list.scss
matthewguywright: Updated.

nickcernis (approved):

PR #420, opened 2022-02-02
[MTKL-1380] Custom Button Components

rfmeier:
## Description

This pull request introduces custom Button components in an attempt to replace hefty styling. The component styling uses [Emotion](https://emotion.sh/docs/introduction) for inline styling instead of predefined styling previously used.

The custom button components consist of `Button`, `TertiaryButton`, `FieldButton`, `LinkButton`, `WarningButton`, and `DarkButton`.


includes/shared-assets/js/components/Buttons.jsx
```diff
@@ -0,0 +1,179 @@
...
+				background: #f4f7fa !important;
+				border-color: #cfdde9 !important;
+				color: #59767f !important;
+			}
+		}
+	`;
+
+	return (
+		<button css={cssAttributes} {...props}>
+			{props.children}
```
rfmeier, line 36: Is `props.children` still a valid way to pass children to a wrapping component?

includes/shared-assets/js/components/Buttons.jsx
```diff
@@ -0,0 +1,179 @@
...
+
+	return (
+		<Button css={cssAttributes} {...props}>
+			{props.children}
+		</Button>
+	);
+}
+
+export function FieldButton(props) {
+	let properties = Object.assign({}, props);
```
rfmeier, line 62: Cloning `props` was needed in order to alter the `className` attribute, but would fail if it did not exist on `props`. Is there a better way to go about this?

includes/shared-assets/js/components/Buttons.jsx
nickcernis: Yep! This is still the way I'd do it.

includes/shared-assets/js/components/Buttons.jsx
```diff
@@ -0,0 +1,179 @@
...
+	`;
+
+	return (
+		<Button css={cssAttributes} {...props}>
+			{props.children}
+		</Button>
+	);
+}
+
+export function FieldButton(props) {
```
nickcernis, line 61: I can't see the `FieldButton` component in use anywhere at the moment.

Is it intended to replace the raw HTML in the `FieldButtons` component?

includes/shared-assets/js/components/Buttons.jsx
nickcernis: Not sure it's better necessarily but this could work:

1. Partially destructure props in the function params so that `className` and `active` are always available even if they aren't passed in (you can optionally set defaults within the destructure syntax if props are missing, as below).
2. Use the `className` as a named attribute of TertiaryButton and spread the other props below that.

That should remove the need to clone/mutate props unless I've missed something.

```js
export function FieldButton({ className = "", active = false, ...props }) {
	let cssAttributes = css`
		align-items: center;
		display: flex;
		margin-right: 8px;
		padding: 10px 16px;

		&:hover svg path,
		&:focus svg path {
			fill: #7e5cef;
		}

		svg {
			margin-right: 9px;
		}
	`;

	if (active) {
		cssAttributes = css`
			${cssAttributes}
			background: #002838;
			color: #fff;

			&:hover,
			&:focus {
				background: #002838;
				border: 1px solid #002838;
				color: #fff;
			}

			svg path,
			&:focus svg path,
			&:hover svg path {
				fill: #fff;
			}
		`;
	}

	return (
		<TertiaryButton
			css={cssAttributes}
			className={`${className} ${active && " active"}`}
			{...props}
		>
			{props.children}
		</TertiaryButton>
	);
}
```

includes/shared-assets/js/components/Buttons.jsx
```diff
@@ -0,0 +1,179 @@
...
+		padding: 10px 16px;
+
+		&:hover svg path,
+		&:focus svg path {
+			fill: #7e5cef;
+		}
+
+		svg {
+			margin-right: 9px;
+		}
```
nickcernis, line 78: Can we remove some of the Sass now this lives here?

For example, this code in `_open-field.scss` looks like it might be redundant now:

```
.field-buttons button {
	align-items: center;
	display: flex;
	margin-right: 8px;
	padding: 10px 16px;
}

.field-buttons button svg {
	margin-right: 9px;
}
```

includes/shared-assets/js/components/Buttons.jsx
nickcernis: Follow-up thought: do we even need the `active` class if active classes are added to the component itself?

rfmeier: Hmm. Maybe this was an oversight on my part.

includes/shared-assets/js/components/Buttons.jsx
rfmeier: This is odd. I had to reimplement it...

includes/shared-assets/js/components/Buttons.jsx
rfmeier: Again with this too. I had to re-remove it.

tests/jest/components/__snapshots__/MediaUploader.test.js.snap
```diff
@@ -13,17 +13,17 @@ exports[`MediaUploader tests Renders a matching snapshot 1`] = `
       className="d-flex flex-row align-items-center media-btns"
     >
       <div>
-        <input
-          className="button button-primary button-large"
-          defaultValue="Upload Media"
+        <DarkButton
+          data-testid="feature-image-button"
```
nickcernis, line 17: Nice, thanks for adding these data-testid attributes.

nickcernis (approved):
Looking great! Nice work.

mindctrl (approved):

PR #425, opened 2022-02-07
[MTKA-1389] Export ACM Blueprints via WP-CLI 

nickcernis:
Adds a `wp acm blueprint export` WP-CLI command to dump the current state of the site as a zipped ACM blueprint.

By default this saves:
- ACM models and fields.
- ACM taxonomies.
- ACM entries.
- WP core posts and pages.
- Taxonomy terms.
- Post tags.
- Post meta.
- Media linked to a post via an ACM media field or stored in `_thumbnail_id` post meta.
- ACM relationships.

https://wpengine.atlassian.net/browse/MTKA-1389

## Testing

Includes integration tests for export functionality.

### To test manually

1. Open a shell ready for WP-CLI. (In Local choose Open Site Shell from the site options menu.)
2. Run `wp acm blueprint export --open`.
3. You should see a directory containing your blueprint zip file. It will contain media if exported posts have post meta for a featured image or ACM media field.

To see all options use `wp help acm blueprint export`.

If you wish you can also try:

1. Taking the zip generated by the export.
2. Uploading it somewhere publicly accessible (such as to a WPE WP test site).
3. Removing existing ACM posts/taxonomies from your site. (`wp option delete atlas_content_modeler_post_types && wp option delete atlas_content_modeler_taxonomies`)
4. Restoring models/taxonomies/other state with `wp acm blueprint import https://example.com/path/to/your/blueprint.zip`.

## Screenshots

Defaults:

```sh
> wp acm blueprint export --open
Collecting ACM data.
Collecting ACM models.
Collecting ACM taxonomies.
Collecting posts.
Collecting terms.
Collecting post meta.
Collecting media.
Collecting ACM relationships
Writing acm.json manifest.
Generating zip.
Opening blueprint temp folder.
Success: Blueprint saved to /var/folders/mc/yf37tb4j5bjbwhv6z0m1bcfm0000gp/T/acm/acm-blueprint/acm-blueprint.zip.
```

With custom options:

```sh
> wp acm blueprint export --name="Custom Name" --description="My blueprint is great." --open
Collecting ACM data.
Collecting ACM models.
Collecting ACM taxonomies.
Collecting posts.
Collecting terms.
Collecting post meta.
Collecting media.
Collecting ACM relationships
Writing acm.json manifest.
Generating zip.
Opening blueprint temp folder.
Success: Blueprint saved to /var/folders/mc/yf37tb4j5bjbwhv6z0m1bcfm0000gp/T/acm/custom-name/custom-name.zip.
```

```sh
[...]

PR #418, opened 2022-02-01
[MTKA-1395] Import ACM Blueprints via WP-CLI

nickcernis:
Adds the ability to import an ACM blueprint using WP-CLI.

https://wpengine.atlassian.net/browse/MTKA-1395

## Testing

Includes integration tests for import functions.

### To test manually

Delete existing ACM models to prevent import conflicts on your first test run.

1. Open a shell ready for WP-CLI (in Local, right-click the site and choose “Open Site Shell”).
2. Run `wp acm blueprint import https://contentmodel1.wpengine.com/wp-content/uploads/2022/02/acm-rabbits.zip`.

Your WP instance should now contain a rabbit model, related taxonomy, media, and rabbit entries with pre-filled data and images.

To reset state for another test (excludes removing posts and media):

```
wp term delete breed cartoon --by=slug && wp term delete breed human --by=slug && wp option delete atlas_content_modeler_taxonomies && wp option delete atlas_content_modeler_post_types
```

To reset only model state (will result in warnings about duplicate taxonomies if importing the same zip file on the next run):

```
wp option delete atlas_content_modeler_post_types
```

To test a blueprint that will fail version checking:

```
wp acm blueprint import http://contentmodel1.wpengine.com/wp-content/uploads/2022/02/acm-needs-newer-version.zip
```

## Screenshots

Successful run (exit 0):

```
> wp acm blueprint import https://contentmodel1.wpengine.com/wp-content/uploads/2022/02/acm-rabbits.zip
Fetching zip.
Unzipping.
Verifying ACM manifest.
Checking minimum versions.
Importing ACM models and fields.
Importing ACM taxonomies.
Importing posts.
Importing terms.
Tagging posts.
Importing media.
Importing post meta.
Restoring ACM relationships.
Deleting zip and manifest.
Success: Import complete.
```

Failed run (fails version check, exit 1):

```
> wp acm blueprint import http://contentmodel1.wpengine.com/wp-content/uploads/2022/02/acm-needs-newer-version.zip
Fetching zip.
Unzipping.
Verifying ACM manifest.
Checking minimum versions.
Error: acm.json requires an ACM version of 100000000000000000000 but the current ACM version is 0.13.0.
```

Failed run (model conflict, exit 1):

```
> wp acm blueprint import https://contentmodel1.wpengine.com/wp-content/uploads/2022/02/acm-rabbits.zip
Cannot load Zend OPcache - it was already loaded
Fetching zip.
Unzipping.
Verifying ACM manifest.
[...]

PR #428, opened 2022-02-10
[NOJIRA] Merge acm blueprint WP-CLI commands to main

nickcernis:
Merges the WP-CLI commands reviewed in #418 and #425 into main to prepare for release.


includes/wp-cli/class-blueprint.php
```diff
@@ -0,0 +1,332 @@
...
+	 * [--min-acm]
+	 * : Minimum Atlas Content Modeler plugin version. Defaults to current
+	 * ACM version.
+	 *
+	 * [--version]
+	 * : Optional blueprint version. Defaults to 1.0.
+	 *
+	 * [--post-types]
+	 * : Post types to include, separated by commas. Defaults to post, page and
+	 * all registered ACM post types.
```
mindctrl, line 226: I just realized that this only affects the post data export and not the model schemas. All model schemas are still included in the export.

` wp acm blueprint export --post-types="rabbit"` gives me an `acm.json` file like this:
```
    "models": {
        "rabbit": {
            ...
        },
        "kitchensink": {
            ...
        }
    },
```

The `posts` property contains only rabbit posts. Should the `--post-types` option also apply to schema data (and taxonomies, terms, etc.)? 

The help text isn't clear on this:
>   [--post-types]
    Post types to include, separated by commas. Defaults to post, page and
    all registered ACM post types.

nickcernis: Thanks for the suggestion, @mindctrl.

I changed the text in https://github.com/wpengine/atlas-content-modeler/pull/428/commits/4c3d2e3944e49abb3961a6dba2a786fbcb52fe98.

I think the flag should only affect posts for now (which also affects post meta/media collection as a side-effect).

anthonyburchell (approved):
Incredible work Nick! With Johns comment addressed I think this looks good. Been looking at it all morning. Let's do it!

mindctrl (approved):

PR #432, opened 2022-02-11
MTKA-1399: update parcel to latest and transformer-sass package to 2.3.1

matthewguywright:
## Description
Update Parcel and transformer sass to fix vulnerability with node-forge.


mindctrl (approved):
LGTM. Everything seems to look and behave as expected. All tests pass. An added bonus is it removes the need for #435

PR #440, opened 2022-02-16
[MTKA-1307] Add minimum and maximum settings to repeatable text fields

rfmeier:
## Description

This pull request adds minimum and maximum settings fields when a text field is marked as repeatable. The publisher will present the repeated fields based on the minimum or maximum settings;
- The minimum number of repeatable text fields will be displayed by default.
- A user is only allowed to delete to the minimum number of repeatable text fields.
- A user can only add to the maximum number of repeatable text fields.

## Screenshots

Screenshot 2022-02-16 at 13-36-03 Models ‹ WordPress Development — WordPress

Screenshot 2022-02-16 at 13-37-37 Add new Rabbit ‹ WordPress Development — WordPress


nickcernis (approved):
Working beautifully — nice job!

Screenshot showing repeatable text fields with different min and max repeatable values.

PR #442, opened 2022-02-22
MTKA-1305: media repeatable

matthewguywright:
## Description


includes/settings/js/src/components/fields/MediaFields.jsx
```diff
@@ -0,0 +1,53 @@
...
+
+const MediaFields = ({ register, data, editing, fieldId }) => {
+	const { models } = useContext(ModelsContext);
+	const query = useLocationSearch();
+	const currentModel = query.get("id");
+	const fields = models[currentModel]?.fields;
+	const [showRepeatable, setShowRepeatable] = useState(
+		data?.isRepeatable === true
+	);
+	const [showTitle, setShowTitle] = useState(data?.isTitle === true);
```
anthonyburchell, line 18: we wont need any concept of title for this one

includes/settings/js/src/components/fields/MediaFields.jsx
```diff
@@ -0,0 +1,53 @@
...
+						{__("Repeatable Field", "atlas-content-modeler")}
+					</legend>
+					<input
+						name="isRepeatable"
+						type="checkbox"
+						id={`is-repeatable-${fieldId}`}
+						ref={register}
+						defaultChecked={showRepeatable}
+						onChange={() => setShowRepeatable(!showRepeatable)}
+						disabled={editing || showTitle}
```
anthonyburchell, line 36: here also you are checking for showtitle and editing

includes/settings/js/src/components/fields/MediaFields.jsx
```diff
@@ -0,0 +1,43 @@
...
+				<div className={"field"}>
+					<legend>
+						{__("Repeatable Field", "atlas-content-modeler")}
+					</legend>
+					<input
+						name="isRepeatable"
+						type="checkbox"
+						id={`is-repeatable-${fieldId}`}
+						ref={register}
+						defaultChecked={showRepeatable}
```
anthonyburchell, line 24: try changing this to `value={showRepeatable}` instead of defaultChecked

includes/publisher/js/src/components/MediaUploader.jsx
```diff
@@ -140,6 +193,283 @@ export default function MediaUploader({ modelSlug, field, required }) {
...
+						url: attachment.attributes.url,
+					});
+				});
+			setValues(imgArr);
+		});
+
+		media.open();
+	}
+
+	if (field.isRepeatable && typeof field.isRepeatable === "boolean") {
```
mindctrl, line 272:
```suggestion
	if (field.isRepeatableMedia && typeof field.isRepeatableMedia === "boolean") {
```

includes/settings/js/src/components/fields/MediaFields.jsx
```diff
@@ -0,0 +1,43 @@
+/**
+ * Additional form fields for the Media field type.
+ */
+import React, { useState } from "react";
+import { __ } from "@wordpress/i18n";
+
+const MediaFields = ({ register, data, editing, fieldId }) => {
+	const [showRepeatable, setShowRepeatable] = useState(
+		data?.isRepeatable === true
```
mindctrl, line 9:
```suggestion
		data?.isRepeatableMedia === true
```

includes/publisher/js/src/components/MediaUploader.jsx
```diff
@@ -140,6 +193,283 @@ export default function MediaUploader({ modelSlug, field, required }) {
...
+																	]{" "}
+																	{getFileName(
+																		item.url
+																	)}
+																</a>
+															</div>
+															<div
+																className={`value[${index}].remove-container p-2 me-sm-1`}
+															>
+																<button
```
mindctrl, line 382: Should this be the TertiaryButton component instead of manually doing the button?

includes/publisher/scss/_media-field.scss
```diff
@@ -0,0 +1,139 @@
+@import "../../../includes/shared-assets/scss/_variables";
+
+.media {
```
mindctrl, line 3: Is there a reason we're doing scss files instead of Emotion here?

mindctrl (changes requested):

includes/publisher/js/src/components/MediaUploader.jsx
```diff
@@ -3,12 +3,23 @@ import { jsx, css } from "@emotion/react";
...
 import {
-	DarkButton,
 	LinkButton,
+	DarkButton,
+	TertiaryButton,
 } from "../../../../shared-assets/js/components/Buttons";
 import { __ } from "@wordpress/i18n";
+import TrashIcon from "../../../../components/icons/TrashIcon";
+import MediaIcon from "../../../../components/icons/MediaIcon";
+
```
mindctrl, line 13: We can use the `acm-icons` alias when importing icons, like we do in other files. However, the TrashIcon is not imported and returned in the [`index.js` file](https://github.com/wpengine/atlas-content-modeler/blob/main/includes/components/icons/index.js), like all the others, so that one would need to be updated to work like the rest.

includes/publisher/js/src/components/MediaUploader.jsx
```diff
@@ -3,12 +3,23 @@ import { jsx, css } from "@emotion/react";
...
+export default function MediaUploader({
+	modelSlug,
+	field,
+	required,
+	errors,
```
mindctrl, line 18:
```suggestion
```

`errors` isn't being used.

includes/publisher/js/src/components/MediaUploader.jsx
```diff
@@ -3,12 +3,23 @@ import { jsx, css } from "@emotion/react";
...
+	defaultError,
```
mindctrl, line 19:
```suggestion
```
```suggestion
```

`errors` and `defaultError` aren't being used.

includes/publisher/js/src/components/MediaUploader.jsx
```diff
@@ -19,30 +30,72 @@ export default function MediaUploader({ modelSlug, field, required }) {
...
 		setValue("");
 		setMediaUrl("");
 	}
 
+	/**
+	 * Sets urls for repeater media field
+	 */
+	function setMultiMediaUrls() {
+		if (Array.isArray(value)) {
+			const imgArr = [];
```
mindctrl, line 60:
```suggestion
```

Not used

includes/publisher/js/src/components/MediaUploader.jsx
```diff
@@ -140,6 +193,283 @@ export default function MediaUploader({ modelSlug, field, required }) {
...
+													<tr
+														key={index}
+														className={`field media-repeater-container-single d-flex mt-1 flex-fill flex-row`}
+													>
+														<td
+															className={`field d-flex flex-row repeater-input mt-0 flex-fill d-lg-flex`}
+														>
+															<div
+																className="me-lg-1 repeater-input-container"
+																name="repeaters"
```
mindctrl, line 305:
```suggestion
```
`name` is an invalid attribute on a `div`

includes/publisher/js/src/components/MediaUploader.jsx
mindctrl: @matthewguywright this `typeof field.isRepeatableMedia === "boolean"` check is what was breaking multi-select for me. The value of `field.isRepeatableMedia` is a string "true" instead of boolean true. Note that other properties use boolean values, so we should update this one to match.

image

includes/publisher/scss/_media-field.scss
matthewguywright: That's the pattern that's in place. It needs to be extracted from the scss and put into the repeatable fields if possible at some point.

includes/publisher/js/src/components/MediaUploader.jsx
```diff
@@ -3,46 +3,112 @@ import { jsx, css } from "@emotion/react";
...
+
 	// state
 	const [mediaUrl, setMediaUrl] = useState("");
 	const [value, setValue] = useState(field.value);
 	const { allowedTypes } = field;
 
 	// local
 	const imageRegex = /\.(gif|jpe?g|tiff?|png|webp|bmp)$/i;
+	const audioRegex = /\.(mp3|ogg|wav|m4a)$/i;
+	const fileRegex = /\.(pdf|doc?x|ppt?x|pps?x|odt|sls?x|psd)$/i;
```
mindctrl, line 22: Where did this list of file extensions come from? What about .txt files? The way this is written, any file type that doesn't match something results in no icon. 

Screenshot
image

What if we refactored it so that if there's no match for a specific type, it defaults to the file icon?

mindctrl:
Just noting we also need some tests here.

includes/publisher/js/src/components/MediaUploader.jsx
matthewguywright: That works. The list comes from the WP allowable upload types for the media modal.

matthewguywright: Txt files are not allowed by WP by default.

Sameedahmad:
media repeatable is really needed, Thank you @matthewguywright  for making it happen.

nickcernis:
@matthewguywright If you create a repeating media field with “use as featured image” ticked and save it on the publisher screen, do you see the error below (“There was an error updating the Array field data”)?

The message seems to be coming from the `if ( 'media' === $field_type` logic starting line 399 of `includes/publisher/class-publisher-form-editing-experience.php`. We probably need to adapt it to account for the fact that the posted values can now be an array of thumbnail IDs and not only a single ID. (Maybe use the first thumbnail ID as the featured image if there's an array of thumbnail IDs?)

Screenshot 2022-03-21 at 15 54 58

includes/components/icons/index.js
```diff
@@ -65,6 +68,14 @@ export default function Icon({ type, size, noCircle }) {
...
 		case "uparrow":
 			return <UpArrow />;
+		case "audio":
+			return <AudioIcon />;
+		case "file":
+			return <FileIcon />;
+		case "multimedia":
+			return <ImgIcon />;
+		case "trash":
+			return <TrashIcon />;
```
andrewbotz, line 78: Trash Icon exists on line 67, should be able to remove one or the other.

mindctrl:
Noting conversation from Slack, when the repeating option is enabled, it does not disable/prevent the enabling of the featured image option — and vice versa.

includes/settings/js/src/components/fields/Form.jsx
```diff
@@ -563,6 +566,7 @@ function Form({ id, position, type, editing, storedData, hasDirtyField }) {
 									defaultChecked={
 										storedData?.required === true
 									}
+									disabled={!!isRepeatableMedia}
```
nickcernis, line 569: It looks like the parent div of this input field needs to change the class if `isRepeatableMedia` is true, so that the label gets ghosted when the checkbox is disabled:

```
							<div
								className={
									!!isRepeatableMedia && "read-only editing"
								}
							>
```

nickcernis, line 569: It looks like the parent div of this input field needs to change the class if `isRepeatableMedia` is true, so that the label gets ghosted when the checkbox is disabled:

```
							<div
								className={
									!!isRepeatableMedia && "read-only editing"
								}
							>
```

includes/settings/js/src/components/fields/MediaFields.jsx
```diff
@@ -0,0 +1,49 @@
...
+	);
+
+	const [isFeatured, setIsFeatured] = useState(data?.isFeatured === true);
+
+	const isFeaturedWatcher = watch("isFeatured");
+
+	return (
+		<>
+			{data && (
+				<div className={"field"}>
```
nickcernis, line 19: Same here — the class should change when the field is not editable so that the label is dimmed.

```suggestion
				<div
					className={
						isFeaturedWatcher ? "field read-only editing" : "field"
					}
				>
```

nickcernis, line 19: Same here — the class should change when the field is not editable so that the label is dimmed.

```suggestion
				<div
					className={
						isFeaturedWatcher ? "field read-only editing" : "field"
					}
				>
```

includes/settings/js/src/components/fields/MediaFields.jsx
```diff
@@ -0,0 +1,49 @@
...
+					<input
+						name="isRepeatableMedia"
+						type="checkbox"
+						id={`is-repeatable-${fieldId}`}
+						ref={register}
+						value={showRepeatableMedia}
+						onChange={() =>
+							setShowRepeatableMedia(!showRepeatableMedia)
+						}
+						disabled={isFeaturedWatcher}
```
nickcernis, line 32: We should probably disable this when it's being edited to be consistent with the other “make repeatable” fields. (You can't change a field's repeatable status once it has been created.)

```suggestion
						disabled={editing || isFeaturedWatcher}
```

tests/acceptance/CreateContentModelMediaFieldCest.php
```diff
@@ -8,6 +8,61 @@ public function _before( \AcceptanceTester $i ) {
...
+		$i->click( '.is-repeatable' );
+		$i->click( '.is-required' );
+		$i->click( 'button[data-testid="edit-model-update-create-button"]' );
+		$i->wait( 1 );
+
+		// Create a new Candies entry.
+		$i->amOnPage( '/wp-admin/edit.php?post_type=candy' );
+		$i->click( 'Add New', '.wrap' );
+		$i->wait( 1 );
+		$i->seeElement( 'input[name="atlas-content-modeler[candy][positionxyz][0]"]:invalid' );
```
nickcernis, line 63: I think this test is failing because no field with a name of `atlas-content-modeler[candy][positionxyz][0]` exists until media has been selected from the media picker. (We only render the repeatable fields once a user has chosen media.)

If you apply the suggestion from Slack to render a hidden required input field when the user has not selected any media, you could then adapt this test to check for that field (by dropping the `[0]`):

```suggestion
		$i->seeElement( 'input[name="atlas-content-modeler[candy][positionxyz]"]:invalid' );
```

tests/acceptance/CreateContentModelMediaFieldCest.php
```diff
@@ -8,6 +8,61 @@ public function _before( \AcceptanceTester $i ) {
...
+		// Create a new Candies entry.
+		$i->amOnPage( '/wp-admin/edit.php?post_type=candy' );
+		$i->click( 'Add New', '.wrap' );
+		$i->wait( 1 );
+
+		// Check that the first input in the index of available list items is rendering indicating a multiple being set.
+		$i->seeElement(
+			'input',
+			[ 'name' => 'atlas-content-modeler[candy][positionxyz][0]' ]
+		);
```
nickcernis, line 39: Same reason for the test failure here, I think: the element you're checking for doesn't exist until media has been selected.

You could check for something that is present instead? For example, instead of `seeElement()`:

```
$i->see( 'Manage Media', 'button[data-testid="media-uploader-manage-media-button"]' );
```

tests/acceptance/CreateContentModelMediaFieldCest.php
```diff
@@ -8,6 +8,58 @@ public function _before( \AcceptanceTester $i ) {
...
+		// Save the field.
+		$i->click( 'button[data-testid="edit-model-update-create-button"]' );
+		$i->wait( 1 );
+
+		// Create a new Candies entry.
+		$i->amOnPage( '/wp-admin/edit.php?post_type=candy' );
+		$i->click( 'Add New', '.wrap' );
+		$i->wait( 1 );
+
+		// Check that the first input in the index of available list items is rendering indicating a multiple being set.
```
nickcernis, line 35: @matthewguywright This comment no longer seems to apply with the change to the line below. Should we delete it?

includes/publisher/js/src/components/MediaUploader.jsx
```diff
@@ -140,6 +211,336 @@ export default function MediaUploader({ modelSlug, field, required }) {
...
+	if (field.isRepeatableMedia) {
+		return (
+			<>
+				<div className={"field"}>
+					<label
+						htmlFor={`atlas-content-modeler[${modelSlug}][${field.slug}]`}
+					>
+						{field.name}
+					</label>
+					{field?.required && <p className="required">*Required</p>}
```
mindctrl, line 299: The "Required" text should be translatable. Something like this (eyeballing this):
```suggestion
					{field?.required && <p className="required">*{__("Required", "atlas-content-modeler")}</p>}
```

mindctrl (changes requested):
GraphQL queries on repeating media are causing WPGraphQL to freeze. The resolver needs an update, around line 541 or so in `custom-post-types-registration.php`.

```
					if ( $field['type'] === [ 'list_of' => 'MediaItem' ] ) {
						return array_map(
							function( $v ) use ( $context ) {
								return DataSource::resolve_post_object( (int) $v, $context );
							},
							$value
						);
					}
```

nickcernis (approved):

mindctrl (approved):

PR #445, opened 2022-02-23
Add/repeatable number field option

anthonyburchell:
## Description
This PR adds settings and publisher side changes that will add repeatable number fields to ACM.


includes/publisher/scss/_number-field.scss
```diff
@@ -0,0 +1,133 @@
+@import "../../../includes/shared-assets/scss/_variables";
+
+.number {
```
nickcernis, line 3: It might be good to put styling like this with the component itself using Emotion instead of adding more SCSS, but we could do it in a separate PR.

includes/publisher/js/src/components/Number.jsx
```diff
@@ -0,0 +1,269 @@
...
+					<p className="help mb-0">{field.description}</p>
+				)}
+				<fieldset>
+					<div id="repeaterNumber" className="text-table flex-row">
+						<div className="repeater-number-field flex-row">
+							<table key="1" className="table mt-2">
+								<tbody>
+									{values.map((item, index) => {
+										return (
+											<>
```
nickcernis, line 76: We can remove the wrapping fragment here — it's only needed if you're returning multiple elements at the top level (we're only returning one <tr> here).

Removing it will helpfully get rid of the “Each child in a list should have a unique "key" prop” console warning (the `key` prop needs to be on the top-level element that's returning from `map`'s callback function — in this case we already have it on`<tr>` so removing the wrapping `<> </>` is enough).

includes/publisher/js/src/components/Number.jsx
```diff
@@ -0,0 +1,269 @@
...
+												<tr
+													key={index}
+													className={`field number-repeater-container-single d-flex mt-1 flex-fill flex-row`}
+												>
+													<div
```
nickcernis, line 81: You can't use a `div` inside a `tr`, unfortunately, which is what this message in the console is warning:

```
Warning: validateDOMNesting(...): <div> cannot appear as a child of <tr>.
```

We'll have to either change the `tr` or the `div`.

includes/publisher/js/src/components/Number.jsx
```diff
@@ -0,0 +1,269 @@
...
+															name="repeaters"
+														>
+															<input
+																ref={
+																	numberInputRef
+																}
+																type={`${field.type}`}
+																name={`atlas-content-modeler[${modelSlug}][${field.slug}][${index}]`}
+																id={`atlas-content-modeler[${modelSlug}][${field.slug}][${index}]`}
+																defaultValue={
```
nickcernis, line 102: Using `defaultValue` and `value` on the same element with React generates this console warning:

> react-dom.js?ver=17.0.1:61 Warning: Number contains an input of type number with both value and defaultValue props. Input elements must be either controlled or uncontrolled (specify either the value prop, or the defaultValue prop, but not both). Decide between using a controlled or uncontrolled input element and remove one of these props. More info: https://reactjs.org/link/controlled-components

You could copy the implementation from the `includes/publisher/js/src/components/Text.jsx` component (which removed `defaultValue` to use just `value` and `onChange`).

anthonyburchell:
Having a bit of a rough time with these e2e tests and jest failures. Will investigate again tomorrow.

includes/publisher/js/src/components/Number.jsx
```diff
@@ -0,0 +1,258 @@
+/* global atlasContentModelerFormEditingExperience */
+import React, { useState, useRef } from "react";
+import Icon from "../../../../components/icons";
+import { __ } from "@wordpress/i18n";
+const { wp } = window;
```
mindctrl, line 5:
```suggestion
```

Appears to be unused.

mindctrl:
The publisher screen isn't rendering for me. I get this error:

```Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.```

**Comment by mindctrl** [comment]
Seeing these warnings:

`<div> cannot appear as a child of <tbody>`

`<button> cannot appear as a child of <tr>`

File: includes/publisher/js/src/components/Number.jsx [hunk 6]
```diff
@@ -0,0 +1,258 @@
...
+						<div className="repeater-number-field flex-row">
+							<table key="1" className="table mt-2">
+								<tbody>
+									{values.map((item, index) => {
+										return (
+											<tr
+												key={index}
+												className={`field number-repeater-container-single d-flex mt-1 flex-fill flex-row`}
+											>
+												<div
```
nickcernis, line 78: Still seeing a warning due to use of div inside tr:

```validateDOMNesting(...): <div> cannot appear as a child of <tr>.```

You could put a `<td>` immediately inside the parent `<tr>` and move all current contents of the `<tr>` into the `<td>` to resolve this, although it may need some styling tweaks.

Comment (anthonyburchell) [reply includes/publisher/js/src/components/Number.jsx:78]: ah yeah. That's what I did in this one and i accounted for the style changes. Not sure why I'm not seeing these warnings on my local but I'll fix it really quick.

Comment (anthonyburchell) [reply includes/publisher/js/src/components/Number.jsx:78]: [fixes table structure and add button markup for hover states](https://github.com/wpengine/atlas-content-modeler/pull/445/commits/55ced17f4ab0f75b0f87608c1095097694b4ec85)

Comment (anthonyburchell) [reply includes/publisher/js/src/components/Number.jsx]: Fixed now in https://github.com/wpengine/atlas-content-modeler/pull/445/commits/b52f955c0e9790d947a579e055bc963fce15d18c

**Review by nickcernis — DISMISSED** [review]
Working great! Nice job! 🎉

File: includes/publisher/js/src/components/Number.jsx [hunk 7]
```diff
@@ -0,0 +1,266 @@
...
+																"repeaterText" +
+																	index
+															]
+																? "field has-error"
+																: "field"
+														} d-flex flex-row repeater-input mt-0 flex-fill d-lg-flex`}
+													>
+														<div
+															className="me-lg-1 repeater-input-container flex-fill"
+															name="repeaters"
```
mindctrl, line 91: My editor flagged `name` as an attribute not allowed on a `div`.

includes/publisher/js/src/components/Number.jsx
```diff
@@ -0,0 +1,266 @@
...
+				event.preventDefault();
+				return;
+			}
+		}
+
+		// call global validate
+		validate(event, field);
+	}
+
+	if (field.isRepeatableNumber) {
```
mindctrl, line 50: Is this going to cause JS errors on existing fields that don't have `isRepeatableNumber` saved in the field config yet?

`field?.isRepeatableNumber` would address that, I think.

includes/settings/js/src/components/fields/NumberFields.jsx
```diff
@@ -1,49 +1,79 @@
...
+				<div className={"field"}>
+					<legend>
+						{__("Repeatable Number Field", "atlas-content-modeler")}
+					</legend>
 					<input
-						type="radio"
-						id="decimal"
-						name="numberType"
-						value="decimal"
+						name="isRepeatableNumber"
```
mindctrl, line 19: Nick briefly explained why you went with `isRepeatableNumber` here instead of `isRepeatable` like the Text field does. It might be worth adding a note here explaining why, for our future selves.

includes/settings/js/src/components/fields/NumberFields.jsx
```diff
@@ -1,49 +1,79 @@
 /**
  * Additional form fields for the Number field type.
  */
-import React from "react";
+import React, { useContext, useState } from "react";
```
mindctrl, line 4:
```suggestion
import React, { useState } from "react";
```

Appears that `useContext` isn't used here.

includes/settings/js/src/components/fields/NumberFields.jsx
```diff
@@ -1,49 +1,79 @@
...
 import { sprintf, __ } from "@wordpress/i18n";
```
mindctrl, line 5:
```suggestion
import { __ } from "@wordpress/i18n";
```

Already existed, but while we're removing the unused import above it...

mindctrl (changes requested):
Left a few small comments, but this is working great! 😎

mindctrl (approved):

PR #446, opened 2022-02-24
MTKA-1370: Save post titles to posts table

### LINKED ISSUE #280: fields used as "entry title" not storing to the database properly, broken search
When I create a model, and chose "Use this field as the entry title" the value of the title is not stored in the database as `post_title` and the Model doesn't implement the `NodeWithTitle` Interface in GraphQL.

### To reproduce
1. Create a model using ACM
2. Create a text field named `title`
3. Select "Use this field as the entry title"
4. Create a post of that model's post_type
5. Enter a value as the title
6. Check the SQL database for the new post and see that `post_title` is "Auto Draft"

### Expected behavior
I expect the value of the field used for "entry title" to be stored as the title. 

Various mechanisms in WordPress, including, but not limited to search, use the values stored in MySQL and if I select "use this field as entry title" I would expect the behaviors of the title to carry over to this field, including things like search behavior. 

### Screenshots

**Create a Model**

Screen Shot 2021-09-21 at 4 00 25 PM

**Field set to "use as entry title"**

Screen Shot 2021-09-21 at 4 01 26 PM

**Post Edit Screen**

✅  Editing the "book" model allows me to save the book with "test" as the value of the "title" field

Screen Shot 2021-09-21 at 3 54 04 PM

**Post List Screen**

✅  The post_list screen shows the title "test" as expected

Screen Shot 2021-09-21 at 3 53 22 PM

**Raw Database Entry**

🚫  The database shows the `post_title` as "Auto Draft"

Screen Shot 2021-09-21 at 4 02 24 PM

**Admin search for the book has no results**

🚫  Searching for the book by title "test" in the admin shows no results

Screen Shot 2021-09-21 at 3 55 19 PM

**Front End Search for the book has no results**

🚫  Searching for the book by title "test" using the default theme's search widget shows no results

Screen Shot 2021-09-21 at 3 58 52 PM

### Version information
- Plugin version: 0.5.0 (I'll go test on a newer version here in a bit too)
- WordPress version: 5.8.1

### LINKED ISSUE #388: Fields set to "use this field as entry title" aren't fully respecting the title responsibilites
This is kind of 2 issues, but both related to creating a field and setting it as "use this field as entry title"

## 1: Slugs aren't generated

When creating a field on a model and marking it as "use this field as the entry title" the behavior doesn't match what the core WordPress title does. Namely, the slug isn't being generated from this field, and as mentioned in #280, the value isn't stored in the `post_title` column in the database.

## 2: Can't name the field "title"

Also, I'm unable to name a field `title` even if I set it to use as the entry title. The client validation prevents this. 

I believe that if I have a field marked as "use this field as the entry title", the validation for using the name "title" should be lifted, and the field should behave as the title. 

### To reproduce
1. Create a new ACM model
2. Add a new text field
3. Name the field "title" and see the validation error that the name can't be used
4. Mark the field "use the field as the entry title" and see the validation error still present. 
5. Try to save the field anyway, and you cannot save it while named `title`, even though you've opted to have the field be the `title`

### Expected behavior
I would expect that if I mark a field "use this field as the entry title" that it would behave like the core title field. It would let me reference it by the name `title`, it would work with existing core methods and filters such as `get_the_title()`, etc. I would expect the title to be used to generate the slug, and I would expect the title to be stored in the `post_title` column and included in search. 

### Screenshots
I recorded a video walking through some things I was experiencing with ACM and this issue is covered in the video.
https://www.youtube.com/watch?v=SmmGqy7bXxo

### Version information
- Plugin version: 0.12.0
- WordPress version: 5.8.2
- Operating system: Mac OSX 11.6.1
- Browser: Chrome v96.0.4664.93

mindctrl:
## Description
This saves title field data to `wp_posts.post_title` instead of `wp_postmeta` like it currently does. As a result, the title filtering logic has been removed.

When fetching a title field's value, if it's in postmeta it is migrated to the posts table and removed from postmeta.

Title data is still available in WPGraphQL as the field name as well as `title`.

Related, the post name/slug is generated from the title when available instead of using the post ID or "entry{id}".

Closes #388 
Closes #280 

https://wpengine.atlassian.net/browse/MTKA-1370

## Testing
Integration tests have been updated to test desired behavior. Please double check the behavior that you see when testing manually.


nickcernis (commented):
The title functionality itself is working well for me!

Screenshot 2022-02-28 at 13 52 10

For the migration logic, one side-effect is that all post titles revert to their original slugs until you visit the post (or cause the title to be fetched in some way).

For example, here's my original Actors admin page index (main branch, actors created with a title text field before checking out this branch):

Screenshot 2022-02-28 at 13 54 15

And here it is as soon as I switch to this PR's branch:

Screenshot 2022-02-28 at 13 54 03

The titles only get filled in once I visit each actor, which I have to do individually:

Screenshot 2022-02-28 at 13 58 05

Not necessarily a blocker but wondering if there's a way we can improve on this? Just thinking about people who might have hundreds of entries that suddenly appear with non-functional titles after this update (in the second screenshot above, how would you know which entry to click to edit Meat Loaf, for example?).

includes/publisher/lib/field-functions.php
```diff
@@ -222,3 +225,205 @@ function sanitize_field( string $type, $value ) {
...
+
+/**
+ * Gets the value of the specified field for the specified post.
+ *
+ * @param array                         $field The field.
+ * @param WP_Post|\WPGraphQL\Model\Post $post The post object.
+ *
+ * @return array|int|mixed|string
+ */
+function get_field_value( array $field, $post ) {
```
mindctrl, line 237: Not sure this is the best place for these functions, but I added them here for now. My original thought was put them in the global namespace and prefix them with `acm_` to make them easier to use. Anyone have thoughts? @wpengine/mintaka 

The reasons for extracting these out of the publisher class was: 1) these are needed outside the publisher class 2) the publisher class is a monster.

nickcernis: I like the idea to extract these — thanks!

Not sure we need to expose them at the top level prefixed by `acm_` yet. I think we should continue to use namespaced functions until we know what we want to expose in a global API.

nickcernis:
@mindctrl New updates are working well for me on the publisher screen!

Screenshot 2022-03-03 at 17 17 30

Just had a question above about whether or not we should delete the title field meta. I think there may be value in preserving it so that it's accessible via the post meta field or post title, even if that means some duplication of data. Or we could filter the title field meta value to return the post title.

andrewbotz (dismissed):

includes/publisher/lib/field-functions.php
```diff
@@ -222,3 +225,205 @@ function sanitize_field( string $type, $value ) {
...
+			 */
+			$meta_value = get_post_meta( $post->ID, $field['slug'], true );
+			if ( ! $meta_value ) {
+				$value = get_post_field( 'post_title', $post->ID );
+				$value = 'Auto Draft' === $value ? '' : $value;
+				break;
+			}
+
+			if ( ! empty( $post->post_title ) || $post->post_title === $meta_value ) {
+				delete_post_meta( $post->ID, $field['slug'] );
```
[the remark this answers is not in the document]
mindctrl: Thanks for catching this. This was a side effect of refactoring the get/save logic. Since some other logic changes `$field['type']` to a WPGraphQL-friendly type, fetching the data in the resolver was triggering the `default` case in the switch — because, e.g., 'String' wasn't a match. See e480179c41a559c074fe3f6662bfe755d76e9ce0. I added a couple tests to check that title fields are also available under their field slug name.

nickcernis (dismissed):
Looking good!

Screenshot 2022-03-14 at 13 44 36
Screenshot 2022-03-14 at 13 44 45

```
> wp post get 1509
Cannot load Zend OPcache - it was already loaded
+-----------------------+------------------------------------------------+
| Field                 | Value                                          |
+-----------------------+------------------------------------------------+
| ID                    | 1509                                           |
| post_author           | 1                                              |
| post_date             | 2022-03-03 16:17:07                            |
| post_date_gmt         | 2022-03-03 16:17:07                            |
| post_content          |                                                |
| post_title            | New Rabbit                                     |
```

anthonyburchell (dismissed):
Looks great! Working well on my end. Big improvement!

includes/publisher/class-publisher-form-editing-experience.php
```diff
@@ -602,48 +520,6 @@ public function render_feedback_notice(): void {
...
-		}
-
-		// Only filter titles for post types created with this plugin.
-		if ( ! array_key_exists( $post_type, $this->models ) ) {
-			return $title;
-		}
-
-		$fields = $this->models[ $post_type ]['fields'] ?? [];
-
-		$title_field = get_entry_title_field( $fields );
```
nickcernis, line 632: Is the `get_entry_title_field` function in use now with this code removal?

I don't see it anywhere except for the tests. Maybe we should remove/deprecate it if so.

luketimothyjones:
Is this issue now being addressed in another branch/PR?

mindctrl:
@luketimothyjones yes we're breaking it down into multiple PRs. Part of that was merged in a feature branch here https://github.com/wpengine/atlas-content-modeler/pull/479

Another PR is coming shortly to address some other parts.

PR #455, opened 2022-03-04
[MTKA-1425] Improve read-only input field styling

nickcernis:
## Description

- Checkbox inputs in the Settings app now appear dimmed only if they are disabled/read-only.
- Disabled inputs such as “Make this field repeatable” use the default instead of the pointer mouse cursor.

https://wpengine.atlassian.net/browse/MTKA-1425

### Checkbox inputs

Confirm checkbox inputs are no longer dimmed if they're clickable:

| Before | After |
| - | - |
| Screenshot 2022-03-04 at 11 08 31 | Screenshot 2022-03-04 at 11 08 18 |

### Disabled inputs
1. Create a new model. 
2. Add a Text field to the model.
3. Check “Use this field as the entry title”.

Hover your mouse over the disabled “Make this field repeatable” checkbox/label. The cursor should be a regular mouse cursor instead of a pointer.

## Documentation Changes

n/a

## Dependant PRs

n/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Fixed
 - The delete model prompt no longer shows “undefined” in its title during model deletion.
+- Improved checkbox styling in field settings.
 
 ## 0.14.0 - 2022-02-10
 ### Added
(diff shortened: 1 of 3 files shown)


andrewbotz (approved):
Everything looks to be working as described.

PR #460, opened 2022-03-14
Fix issue where fatal errors happen on leading number model names

anthonyburchell:
This PR uses the toValidApiId format for generating a slug. Previously using a leading number in a slug name would cause fatal errors on the site. toValidAPiId factors in this limitation.

### DIFF
--- a/includes/settings/js/src/components/CreateContentModel.jsx
+++ b/includes/settings/js/src/components/CreateContentModel.jsx
@@ -260,7 +260,6 @@ export default function CreateContentModel() {
 							ref={register({ required: true, maxLength: 20 })}
 							onChange={(event) => {
 								onChangeGeneratedValue(event.target.value);
-								toValidApiId(event.target.value);
 							}}
 						/>
 						<p className="field-messages">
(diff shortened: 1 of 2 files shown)


includes/settings/js/src/components/CreateContentModel.jsx
```diff
@@ -5,7 +5,7 @@ import { useHistory } from "react-router-dom";
 import { ModelsContext } from "../ModelsContext";
 import { insertSidebarMenuItem } from "../utils";
 import { useInputGenerator } from "../hooks";
-import { toSanitizedKey } from "../formats";
+import { toSanitizedKey, toValidApiId } from "../formats";
```
nickcernis, line 8: It looks like `toSanitizedKey` is an unused import now, and could be removed here.

```suggestion
import { toValidApiId } from "../formats";
```

nickcernis (approved):
Working well for me — I tested auto-generated model slugs and manually editing the slug and wasn't able to introduce a leading number either way.

Just one comment about an unused import, and the changelog conflict to fix up!

nickcernis (changes requested):
In Slack we found that the name and plural names also need leading numbers to be stripped. Changing my review so we don't merge in error.

anthonyburchell:
closing in favor of: https://github.com/wpengine/atlas-content-modeler/pull/463

PR #462, opened 2022-03-15
MTKA-1302: Add Rich Text repeatable option

nickcernis:
## Description

Rich Text fields now include an optional “Make this field repeatable” checkbox.

It behaves the same way as the Text field's repeatable option, allowing publishers to input multiple rich text fields as separate rows.

https://wpengine.atlassian.net/browse/mtka-1302

## Testing

Includes e2e tests, unit tests for GraphQL/REST, and Jest tests for components.

To test manually:

You may need to first `npm ci` before you `npm start` or `npm run build`. I added `uuid` as a dependency (it was already a sub-dependency of Jest; this just allows us to use safely in front-end code).

1. Create a model with the rich text field.
2. Tick the “make repeatable” option.
3. Add an entry, then enter and save repeatable rich text content.

### Developer side

rich-text-option-developer-app

### Publisher side

rich-text-repeater-single-field

rich-text-repeater-multiple-fields

## Documentation Changes

n/a 

## Dependant PRs

n/a


includes/shared-assets/js/emotion.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Shared styles for components styled with Emotion.
+ */
+
+export const colors = {
```
andrewbotz, line 5: Love this change and the use of colors object throughout. I was wondering if we could use emotion's ThemeProvider that way we can get all these values plus anything else without having to ever worry about importing colors.

Here's an example : https://github.com/wpengine/atlas-content-modeler/pull/456/files 

Ultimately we could do this for colors, but also common values in relation to other things, such as typography and the alike.

includes/publisher/js/src/components/RichText/SoloRichTextEditorField.jsx
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import RichTextEditorHeader from "./RichTextEditorHeader";
+
+const SoloRichTextEditorField = ({ modelSlug, field, fieldId }) => {
```
andrewbotz, line 4: Clean! Love it.

includes/publisher/js/src/components/RichText/RichTextEditor.jsx
```diff
@@ -0,0 +1,76 @@
...
+		: [{ id: uuidv4(), value: field?.value }];
+
+	const [values, setValues] = useState(initialValues);
+
+	useEffect(() => {
+		const editorReadyTime = 500;
+		const initializeEditorWhenReady = () => {
+			/**
+			 * WP defines getDefaultSettings() in an admin footer script tag after
+			 * admin scripts are enqueued, so we must wait for it to be available.
```
andrewbotz, line 24: Love the comments. Curious, do we need to do this often? As in, set a timeout for WP to finish something?

includes/shared-assets/js/emotion.js
nickcernis: I really like this idea! I hadn't considered using ThemeProvider because we have no plans to offer styling variations just yet, but the API looks clean and it could help us in the future if we need to implement a dark mode etc.

Not sure if it belongs in this PR or is better done as a separate story, though. I went with importing a bare object here just because it's pretty minimal. Open to ideas on that, though. Perhaps I could pull out my color imports and just use bare strings, and we can refactor later?

includes/publisher/js/src/components/RichText/RichTextEditor.jsx
nickcernis: It's pretty rare! (I think this might be the only place we do it.)

We can usually set a WP core script as a dependency of our script, which ensures it loads before our script has loaded. With scripts that WP core has injected inline, that becomes harder, which is why I went with the workaround here. More details into what it solves and how here: https://github.com/wpengine/atlas-content-modeler/pull/400

rfmeier:
Your buttons are nice and clean. Would it be worth moving them to a more generic directory? They can easily be reused with the other repeatable fields.

nickcernis:
@rfmeier Sure, I can do that. I think they'll need some light refactoring — we should probably pass an `addItem` and `removeItem` function prop to each (instead of `setValues`) if we want to use them more generically.

That way the button would receive a function that it passes straight to `onClick` instead of holding logic to add/remove rows itself. (I had to adapt the Rich Text field to use uuids for each field instead of a 0-based index, which might not be needed for other fields, so the buttons in this PR currently include uuid logic that others might not need.)

Will take a look tomorrow…

includes/shared-assets/js/emotion.js
andrewbotz: Oh ya, totally fine not diving into ThemeProvider now.

includes/publisher/js/src/components/RichText/RichTextEditor.jsx
andrewbotz: Thank you for the reply and explanation.

nickcernis:
@rfmeier I refactored the buttons in https://github.com/wpengine/atlas-content-modeler/pull/462/commits/c6d5219b62aed907cc1f20087233e72a3ce05304 and moved them in https://github.com/wpengine/atlas-content-modeler/pull/462/commits/0abdff55d2783ea90a77deaf43bf1a95941a06bc.

Nice idea to make them more reuseable. If you see other ways to improve them feel free to say.

rfmeier (approved):
Excellent. I didn't see any issues. Great work 👍

mindctrl (approved):

PR #463, opened 2022-03-15
MTKA-1426 adds errors for numerical first letter characters in slug, plural, and singular model ids

anthonyburchell:
## Description
@bhardie had the idea of validating leading characters the same way we do it with validating that a model name exists or is empty. This is consistent with our error handling and also prevents the fatal errors introduced when creating a model with a leading numerical character.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Improved checkbox styling in field settings.
 - Options buttons now use the mouse pointer cursor.
 - The `wp acm blueprint import` command no longer reports “Could not read an acm.json file” if the blueprint zip file was renamed after creation.
+- Fixed issue where it was possible to improperly lead a model ID with a number.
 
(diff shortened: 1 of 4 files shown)


andrewbotz (dismissed):
Works great. Tested all three inputs and was unable to use leading numeric value.

nickcernis (commented):
Working well! A couple of things:

- I'm able to enter a model singular or plural name with a leading space followed by a number and it passes validation (but fails in GraphQL). For example, try entering " 123Test". It looks like we need to run `trim()` on these values before we validate against them or store them.
- Do you think it's worth adding tests for the new errors you added here? `tests/integration/api-validation/test-rest-model-endpoints.php` might be a good place to extend.

mindctrl (approved):
Noted in Slack there are still some edge cases, but we'll address those in a separate issue. This fixes the fatal error issue as described.

PR #468, opened 2022-03-22
MTKA-1452: Let models override with_front

nickcernis:
## Description

Addresses #457 by adding a “Use Permalink Base” option on the model edit screen to override the with_front setting when registering the model's post type.

https://wpengine.atlassian.net/browse/MTKA-1452

## Details / Context

WordPress registers post types with `'rewrite' => [ 'with_front' => true ]` by default.

This means a site using a custom permalink structure of `/posts/%postname%/` sees their ACM post URLs prefixed by `/posts/`: `/posts/your-acm-model-name/your-post/`.

A user can now untick “Use Permalink Base” if they want `/posts/%postname%/` for regular WP posts, but prefer ACM posts to appear at `/your-acm-model-name/your-post/` (without the `/posts/` prefix).

The PR handles this in a way that does not change behavior for existing models (they will continue to use `with_front` unless a user overrides them by editing the model).

This also does not change the default behavior for new models (new models have empty `with_front`, but custom post type registration logic sets it to true if `with_front` is absent). This is consistent with original behavior and WP defaults.

I chose not to add the field to the "create new model" form, as it's not something a user will need to set or change often, and it reduces the size of the create new model form. (i.e. You can only access the setting when editing a model, not when creating one.) Happy to add it there too if desired, though.

## Testing

Includes e2e tests to check field interactivity and unit tests to check `with_front` applies.

### To test manually

1. Set Permalinks to this custom structure at Settings → Permalinks: `/posts/%postname%/`
2. Create a new model with one field and at least one entry.
3. Run a GraphQL query, being sure to include `uri` as one of the field properties (see screenshots below). The uri in the query response should include the `/posts/` prefix by default.
4. Edit the model and untick the new `“Use Permalink Base”` option.
5. Re-run the query. `/posts/` should no longer appear in the uri.

## Screenshots

Screenshot 2022-03-22 at 18 03 11

Screenshot 2022-03-22 at 18 00 37

Screenshot 2022-03-22 at 18 02 24

Screenshot 2022-03-22 at 18 03 41

Screenshot 2022-03-22 at 18 04 12

## Documentation Changes

n/a

## Dependant PRs

n/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 ### Added
 - Rich Text fields have a new “make repeatable” option. Enable it on new fields to let publishers enter multiple rows of rich text content.
+- Added a “Use Permalink Base” option on the edit model screen to set the WordPress `with_front` setting (true by default). Untick this to tell WordPress not to prefix your model entry URLs with custom prefixes from your WordPress permalink settings. For example, a site with a permalink structure of `/posts/%postname%/` will have post URLs of `/posts/your-acm-model-name/your-post/` by default. Edit your model and untick “Use Permalink Base” if you prefer a URL structure of `/your-acm-model-name/your-post/`.
 - `wp acm blueprint export` now accepts a `--wp-options` flag to export a comma-separated list of WordPress options from the `wp_options` table. No options are exported by default. Example: `wp acm blueprint export --wp-options='blogname, permalink_structure`
 - `wp acm blueprint import` now updates WordPress options if blueprints contain a `wp-options` key with a list of options values, keyed by option name.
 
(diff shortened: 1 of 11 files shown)


mindctrl (dismissed):
This works great! The only thing I noticed that might confuse someone is when you change this setting, navigate to the Rabbits list table and view a post, it returns a 404 because the permalinks aren't flushed.

nickcernis:
@mindctrl Thanks for the review and note! Tracked the rewrite flush thing down to update_registered_content_types(). Should be fixed in https://github.com/wpengine/atlas-content-modeler/pull/468/commits/bf98470c52b5665b7420d002bc4b70fc5ca51c13.

mindctrl (dismissed):
💥

nickcernis:
Now ready for final review and merge. Sorry for the review churn.

mindctrl (approved):

PR #474, opened 2022-03-24
Feature/publisher repeatable date fields

anthonyburchell:
## Description

This PR adds the publisher side changes to allow for repeatable date fields. To test, create a date field and mark "repeatable"

From there you'll be able to add a date repeater where you can define an array of dates for your entries.


includes/publisher/js/src/components/Date/RepeatingDate.jsx
```diff
@@ -0,0 +1,96 @@
...
+					width: 100%;
+					max-width: 60%;
+					margin-top: 15px;
+					border-spacing: 0;
+					border: solid 1px ${colors.border};
+				`}
+			>
+				{values.map(({ id, value }, index) => {
+					return (
+						<tr key={id} data-testid="date-repeater-row">
```
nickcernis, line 39: I'm seeing “Each child in a list should have a unique "key" prop.” in the console in relation to this code.

It's because there is no `id` value in 'values', so ids are all undefined and not unique. (See `initialValues` in Date.jsx — it only has a 'value' property, no 'id'. I used 'id' in the Rich Text repeater to create a unique ID for each field, which was required there to initialize multiple TinyMCE instances, but that code isn't present in the Date field.)

This means we can't use 'id' in the destructuring code or in the key prop.

You could remove the id, then use the index as the key:

```suggestion
				{values.map(({ value }, index) => {
					return (
						<tr key={index} data-testid="date-repeater-row">
```

includes/publisher/js/src/components/Date/RepeatingDate.jsx
```diff
@@ -0,0 +1,96 @@
...
+	const deleteItem = (index) =>
+		setValues((currentValues) => {
+			const newValues = [...currentValues];
+			newValues.splice(index, 1);
+			return newValues;
+		});
+
+	return (
+		<fieldset>
+			<table
```
nickcernis, line 27: Looks like we need a `<tbody>` here inside `<table>` (and tweaking the styles too). 

Without it I see this in the console:

> <tr> cannot appear as a child of <table>. Add a <tbody>, <thead> or <tfoot> to your code to match the DOM tree generated by the browser.

includes/publisher/js/src/components/Date/RepeatingDate.jsx
```diff
@@ -0,0 +1,96 @@
...
+						</tr>
+					);
+				})}
+				<tr>
+					<td>
+						<AddItemButton
+							data-testid={`add-date-field-button`}
+							addItem={addItem}
+							css={css`
+								margin: 0 !important;
```
nickcernis, line 86: This works fine without the `!important` for me (Emotion adds a second `margin`, which overrides the first).

```suggestion
								margin: 0;
```

nickcernis (commented):
@anthonyburchell Looking great!

Just a couple of small things above to fix some console errors, then should be good to merge. 🥳

nickcernis:
@anthonyburchell Please could you also add a changelog entry to unreleased? Thanks!

nickcernis (approved):

PR #479, opened 2022-03-30
Save post title data to wp_posts table

mindctrl:
## Description
Saves post title data to `wp_posts` table when a title field is saved to `wp_postmeta`.

More title work coming in other PRs.

https://wpengine.atlassian.net/browse/MTKA-1370

## Testing
- Add a model with a Text field configured to be the Title Field
- Add a new post entry
- Confirm that the values you enter into the Text field are saved to the post_title field of the new post. (postmeta is left as is for now)
- Update post, confirm update is saved to post_title field.


nickcernis (approved):
Looking good! I like the approach of copying the post title on meta change.

```
> wp post meta get 1777 name
Fred

> wp post get 1777 | grep post_title
post_title	Fred
```

PR #482, opened 2022-03-31
MTKA-1438: Add email type as string to graphql

andrewbotz:
## Description

Allow email type to be queryable via GraphQL.

https://wpengine.atlassian.net/browse/MTKA-1438

## Testing

1. Create new content model that includes an email
2. Publish model with an input type of email
3. Attempt to query published data model via GraphQL

## Screenshots

Screen Shot 2022-03-31 at 11 18 42 AM
Screen Shot 2022-03-31 at 11 18 33 AM
Screen Shot 2022-03-31 at 11 15 39 AM

### DIFF
--- a/includes/content-registration/custom-post-types-registration.php
+++ b/includes/content-registration/custom-post-types-registration.php
@@ -760,6 +760,7 @@ function map_html_field_type_to_graphql_field_type( string $field_type ): ?strin
 		case 'string':
 		case 'date':
 		case 'richtext':
+		case 'email':
 			return 'String';
 		case 'multipleChoice':
 			return 'list';
--- a/tests/integration/api-validation/test-data/fields.php
+++ b/tests/integration/api-validation/test-data/fields.php
@@ -411,5 +411,18 @@ function get_test_fields() {
 			'minChars'         => '',
 			'maxChars'         => '',
 		),
+		'1648828035'    => array(
+			'show_in_rest'     => true,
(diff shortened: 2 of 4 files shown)


rfmeier (approved):

mindctrl (approved):
LGTM. Works as expected. Noting that the repeating stuff is coming in another PR. Even though this field is very similar to a text field as far as GraphQL is concerned, it would be nice to have some tests that confirm this field is in the GraphQL response as expected to prevent regressions.

nickcernis:
@andrewbotz This is looking good! Some pointers in case you want to add the GraphQL tests @mindctrl mentions to check that we're getting GraphQL response data for email fields. You would need to:

1. Extend `get_test_fields` in `tests/integration/api-validation/test-data/fields.php` with an email field.
2. Populate that email field with an email address in `populate_post` in `tests/integration/api-validation/test-data/posts.php` (via a new `update_post_meta` call near the bottom of that function, and using the 'slug' you set for the field in step 1).
3. Modify `test_graphql_query_result_has_custom_fields_data` in `tests/integration/api-validation/test-graphql-model-data.php` to add the email field slug in the query (the list of fields under the 'nodes' property), and to check the response for the email field content (the `self::assertArrayHasKey` assertations below — you can check there for the email address you set in step 2).

If you want a walkthrough over Slack/Zoom, feel free to reach out.

PR #488, opened 2022-04-01
add crud fetch model function

anthonyburchell:
This PR adds a fetch_model function to our api classes for the crud functionality.


includes/api/crud-functions.php
```diff
@@ -9,7 +9,86 @@
...
+ *
+ * @return bool|WP_Error False or WP_Error if relation could not be made, else true.
+ */
+function add_relationship( int $post_id, string $relationship_field_slug, int $relationship_id ) {
+	$post = get_post( $post_id );
+	if ( empty( $post ) ) {
+		return new \WP_Error( 'invalid_post_object', 'The post object was invalid' );
+	}
+
+	$field = get_field_from_slug( $relationship_field_slug, get_option( 'atlas_content_modeler_post_types' ), $post->post_type );
```
mindctrl, line 51:
```suggestion
	$field = get_field_from_slug( $relationship_field_slug, get_registered_content_types(), $post->post_type );
```

includes/api/crud-functions.php
```diff
@@ -9,7 +9,86 @@
...
+ *
+ * @return bool|WP_Error False or WP_Error if relation could not be made, else true.
+ */
+function replace_relationship( int $post_id, string $relationship_field_slug, array $relationship_ids ) {
+	$post = get_post( $post_id );
+	if ( empty( $post ) ) {
+		return new \WP_Error( 'invalid_post_object', 'The post object was invalid' );
+	}
+
+	$field = get_field_from_slug( $relationship_field_slug, get_option( 'atlas_content_modeler_post_types' ), $post->post_type );
```
mindctrl, line 82:
```suggestion
	$field = get_field_from_slug( $relationship_field_slug, get_registered_content_types(), $post->post_type );
```

mindctrl (changes requested):

includes/api/crud-functions.php
anthonyburchell: @rfmeier fyi ^

I think this code got merged in a strange way as Ryan and I were working separately toward the feature branch. this was merged but I agree with the change. How should we best do this Ryan? okay to include in this PR?

mindctrl: If it's annoying to do now, we can do this in another branch or during the final PR review

mindctrl:
There's a merge conflict here. I think if you resolve that, the diff view will look a lot cleaner and only show what you're adding in this PR.

includes/api/crud-functions.php
rfmeier: Either way works. A follow-up PR shouldn't require much effort.

includes/api/crud-functions.php
```diff
@@ -10,12 +10,28 @@
...
+/**
+ * Gets the model data for a defined content model.
+ *
+ * @param string $model The content model slug.
+ * @return array
+ */
+function fetch_model( string $model ): array {
+	$models = get_registered_content_types();
+	if ( empty( $models[ $model ] ) ) {
+		return [];
```
rfmeier, line 30: Should the invalid return be `null` instead of an array? I know the model is an array, but technically an object. Arrays usually imply a list of.

rfmeier (changes requested):

includes/api/crud-functions.php
anthonyburchell: good call. done.

rfmeier (approved):

mindctrl (approved):

PR #492, opened 2022-04-06
Mtka 1439 dev email domain validation

andrewbotz:
## Description

Add in Advanced Settings for Email that allows for validating Domains and Subdomains. This PR does not include the Publisher side, it only wires up the Developer side.

https://wpengine.atlassian.net/browse/MTKA-1439

## Testing

1. Create new Content Model with an Email Input field.
2. Click on Advanced Settings
3. Set a domain and/or subdomain
4. Attempt to create new model with Email input

## Screenshots

Screen Shot 2022-04-06 at 4 08 13 PM
Screen Shot 2022-04-06 at 4 08 22 PM
Screen Shot 2022-04-06 at 4 08 31 PM
Screen Shot 2022-04-06 at 4 08 50 PM

### DIFF
--- a/includes/settings/js/src/components/fields/EmailSettings.jsx
+++ b/includes/settings/js/src/components/fields/EmailSettings.jsx
@@ -4,20 +4,14 @@ import React from "react";
 import Icon from "../../../../../components/icons";
 import { __ } from "@wordpress/i18n";
 
-// TODO-abotz: figure out why we need important here and remove
-const emailDomainTextArea = css({
-	maxWidth: "none !important",
-	minHeight: "55px !important",
-});
+const emailDomainTextArea = css`
+	.atlas-content-modeler-admin-page & {
+		max-width: none;
+		min-height: 55px;
+	}
+`;
 
-const EmailSettings = ({
-	errors,
-	field,
-	storedData,
-	setValue,
-	getValues,
-	trigger,
-}) => {
+const EmailSettings = ({ errors, storedData, setValue, getValues }) => {
 	return (
 		<>
 			<div className="d-flex flex-column d-sm-flex flex-sm-row">
@@ -35,7 +29,7 @@ const EmailSettings = ({
 						</label>
 						<p>
 							{__(
-								`Comma seperated list of allowed domains, wild cards allowed. Eg "*.edu"`,
+								`Comma-separated list of allowed domains, wildcards allowed. e.g. "*.edu"`,
 								"atlas-content-modeler"
 							)}
 						</p>
--- a/includes/settings/js/src/components/fields/EmailSettings.jsx
+++ b/includes/settings/js/src/components/fields/EmailSettings.jsx
@@ -1,0 +1,90 @@
+/** @jsx jsx */
+import { jsx, css } from "@emotion/react";
+import React from "react";
+import Icon from "../../../../../components/icons";
+import { __ } from "@wordpress/i18n";
+
+// TODO-abotz: figure out why we need important here and remove
+const emailDomainTextArea = css({
+	maxWidth: "none !important",
+	minHeight: "55px !important",
+});
+
+const EmailSettings = ({
+	errors,
+	field,
+	storedData,
+	setValue,
+	getValues,
+	trigger,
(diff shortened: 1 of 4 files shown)


includes/settings/js/src/components/fields/EmailSettings.jsx
```diff
@@ -0,0 +1,90 @@
...
+					>
+						<label htmlFor="allowedDomains">
+							{__(
+								"Specify Allowed Domains",
+								"atlas-content-modeler"
+							)}
+						</label>
+						<p>
+							{__(
+								`Comma seperated list of allowed domains, wild cards allowed. Eg "*.edu"`,
```
nickcernis, line 38: Nit: spelling/grammar

```suggestion
								`Comma-separated list of allowed domains, wildcards allowed. e.g. "*.edu"`,
```

includes/settings/js/src/components/fields/Form.jsx
```diff
@@ -139,6 +140,32 @@ function Form({ id, position, type, editing, storedData, hasDirtyField }) {
...
+						value ? value.replace(/\s/g, "") : "",
+					validate: {
+						formattedCorrectly: (value) => {
+							if (!value) {
+								return true;
+							}
+
+							const domains = value
+								.split(",")
+								.map((domain) => domain.trim());
```
nickcernis, line 157: Is it worth stripping empty strings here so that lists like `example.net, example.com,` (trailing comma at end of list) and `example.com, ,test.example.com` (double comma) pass validation?

```suggestion
							const domains = value
								.split(",")
								.map((domain) => domain.trim())
								.filter((domain) => domain.length > 0);
```

includes/settings/js/src/components/fields/EmailSettings.jsx
```diff
@@ -0,0 +1,90 @@
...
+	minHeight: "55px !important",
+});
+
+const EmailSettings = ({
+	errors,
+	field,
+	storedData,
+	setValue,
+	getValues,
+	trigger,
```
nickcernis, line 19: It looks like `field` and `trigger` are not being used and could be removed here.

includes/settings/js/src/components/fields/EmailSettings.jsx
```diff
@@ -0,0 +1,90 @@
...
+import { jsx, css } from "@emotion/react";
+import React from "react";
+import Icon from "../../../../../components/icons";
+import { __ } from "@wordpress/i18n";
+
+// TODO-abotz: figure out why we need important here and remove
+const emailDomainTextArea = css({
+	maxWidth: "none !important",
+	minHeight: "55px !important",
+});
```
nickcernis, line 11: Inline Emotion styles will be overridden by more specific external CSS by default (in this case the `.atlas-content-modeler-admin-page textarea` styles in our `_forms.scss` are overriding the Emotion styles).

To work around this you could use the `css` JS template string and write your CSS as normal but with a `.atlas-content-modeler-admin-page` prefix:

```suggestion
const emailDomainTextArea = css`
	.atlas-content-modeler-admin-page & {
		max-width: none;
		min-height: 55px;
	}
`;
```

The `&` there will be replaced by the class that Emotion generates automatically, resulting in a CSS rule like this that overrides the `.atlas-content-modeler-admin-page textarea` styling:

```css
.atlas-content-modeler-admin-page .css-bkv9i3-EmailSettings {
    max-width: none;
    min-height: 55px;
}
```

It's not great that we have to do this in some cases, but it could be reduced over time if we cut down on generic/global Sass and CSS.

includes/settings/js/src/components/fields/Form.jsx
andrewbotz: Nice! Will change.

includes/settings/js/src/components/fields/EmailSettings.jsx
andrewbotz: Thanks for the explanation man, makes sense!

nickcernis (approved):

mindctrl (approved):

PR #495, opened 2022-04-08
chore: update post title when fetching title field data from post meta.

mindctrl:
## Description

Updates post title in posts table when title field value is fetched from post meta.

Part of https://wpengine.atlassian.net/browse/MTKA-1370

## Testing

Automated tests should continue to pass.


nickcernis (approved):
Working well for me!

It looks like this copies the title field to post_title when the user visits the publisher page for each entry. They'd have to do that manually for all entries to migrate them all. Is that the migration behavior we want, or just the best we can hope to have without a full migration process?

nickcernis (approved):

PR #500, opened 2022-04-14
MTKA-1469 Email Domain/Subdomain Validation for Publisher

andrewbotz:
## Description

Publisher Email input validation with Domain and Subdomain criteria.

Parent Ticket : https://wpengine.atlassian.net/browse/MTKA-1439
Subtask Ticket : https://wpengine.atlassian.net/browse/MTKA-1469

## Testing

- Create new model with Email input
- Add Advanced Settings -> Domain/Subdomain criteria
- Go to Publisher using said model
- Attempt to Publish model that includes an email input field with Advanced Settings -> Domain/Subdomain criteria
- Affirm cannot publish email input field if it does not meet domain/subdomain criteria

## Screenshots

Screen Shot 2022-04-14 at 1 48 34 PM
Screen Shot 2022-04-14 at 1 48 49 PM
Screen Shot 2022-04-14 at 1 49 04 PM
Screen Shot 2022-04-14 at 1 49 14 PM
Screen Shot 2022-04-14 at 1 49 27 PM

### DIFF
--- a/includes/publisher/js/src/components/Email.jsx
+++ b/includes/publisher/js/src/components/Email.jsx
@@ -1,6 +1,7 @@
 import React, { useState } from "react";
 import { __ } from "@wordpress/i18n";
 import Icon from "../../../../components/icons";
+import { buildWildcardRegex } from "../../../../shared-assets/js/validation/emailValidation";
 
 export default function Email({
 	field,
@@ -16,6 +17,7 @@ export default function Email({
 		defaultValue: field.value,
 		required: field.required,
 		onChange: (event) => validate(event, field),
+		pattern: buildWildcardRegex(field.allowedDomains),
 	};
 
 	return (
--- a/includes/publisher/js/src/components/Email.jsx
+++ b/includes/publisher/js/src/components/Email.jsx
@@ -33,6 +33,13 @@ export default function Email({
 			{field?.description && (
 				<p className="help mb-0">{field.description}</p>
 			)}
+			{field?.allowedDomains && (
+				<p className="help mb-0">
+					Allowed domains :{" "}
+					{field.allowedDomains.replaceAll(",", ", ")}
+				</p>
+			)}
+
 			<input {...emailProps} />
 			<span className="error">
 				<Icon type="error" />
--- a/includes/publisher/js/src/components/Field.jsx
+++ b/includes/publisher/js/src/components/Field.jsx
@@ -8,7 +8,7 @@ import Number from "./Number";
 import Date from "./Date";
 import Icon from "acm-icons";
 import { sprintf, __ } from "@wordpress/i18n";
-import { isValidDomain } from "../../../../shared-assets/js/validation/emailValidation";
+import { isEmailDomainValid } from "../../../../shared-assets/js/validation/emailValidation";
 
 const defaultError = "This field is required";
 const integerRegex = /^[-+]?\d+/g;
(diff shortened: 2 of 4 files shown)


includes/publisher/js/src/components/Field.jsx
```diff
@@ -87,6 +88,16 @@ export default function Field(props) {
...
+
+			if (
+				field?.allowedDomains &&
+				event.target.validity.patternMismatch
+			) {
+				error = __(
+					"Email must include domain or subdomain criteria.",
+					"atlas-content-modeler"
+				);
+			}
```
nickcernis, line 100: Could this message be clearer? Perhaps:

```suggestion
				error = __(
					"Email must end with an allowed domain.",
					"atlas-content-modeler"
				);
			}
```

(Just mirroring the "allowed domains" that you use above the field, which is great, and perhaps more obvious than “…domain criteria”.)

includes/publisher/js/src/components/Email.jsx
```diff
@@ -16,6 +17,7 @@ export default function Email({
 		defaultValue: field.value,
 		required: field.required,
 		onChange: (event) => validate(event, field),
+		pattern: buildWildcardRegex(field.allowedDomains),
```
nickcernis, line 20: Worth setting a fallback here in case `field.allowedDomains` is `undefined` for some reason?

```suggestion
		pattern: buildWildcardRegex(field.allowedDomains ?? ""),
```

I still had some email fields from before you added the `allowedDomains` property, which causes an empty form to display due to `values.split` failing in `buildWildcardRegex` (see below). A fallback might protect us from similar issues if `allowedDomains` isn't saved for some other reason. But feel free to ignore if you think it's overly defensive.

Screenshot 2022-04-15 at 09 44 22

tests/acceptance/PublishModelCest.php
```diff
@@ -195,7 +200,7 @@ public function i_see_submission_errors_in_email_field_when_input_is_not_an_emai
 		$i->click( 'Publish', '#publishing-action' );
 		$i->wait( 2 );
 
-		$i->see( 'Value must be an email.' );
+		$i->see( 'Email must include domain or subdomain criteria.' );
```
mindctrl, line 203: It'd be nice to see a test that confirms the behavioral difference between things like `*.wpengine.com` and `wpengine.com`

includes/publisher/js/src/components/Email.jsx
andrewbotz: I like this, in fact I had some defensive checks with the `buildWildcardRegex(...)` but removed unfortunately. May I make a suggestion though, instead of setting the default pattern  value to be `""` from the component(s), could I set default value within the `buildWildcardRegex(...)` function? 

Something like this :

```JavaScript
// ... brevity
export const buildWildcardRegex = (values = "", delimiter = ",") => {
// ...
```

nickcernis: Sounds good, nice idea!

tests/acceptance/PublishModelCest.php
andrewbotz: Of course, will do! 

I also want to add some tests on the emailValidation.js file as well -- wanted to get this out into PR before weekend just in case I needed to refactor/change anything major. Thanks for the feedback as always.

mindctrl (approved):

PR #502, opened 2022-04-19
[MTKA-1479] refactor post slug generation

mindctrl:
## Description

- [x] Continue generating fallback slug on posts with no title field
- [x] Generate post slug from title field
- [x] Tests

https://wpengine.atlassian.net/browse/MTKA-1479

### DIFF
--- a/includes/publisher/class-publisher-form-editing-experience.php
+++ b/includes/publisher/class-publisher-form-editing-experience.php
@@ -315,6 +315,15 @@ final class FormEditingExperience {
 			return;
 		}
 
+		if ( $post->post_status === 'auto-draft' ) {
+			$post->post_name  = $post->ID;
+			$post->post_title = 'entry' . $post->ID;
+			remove_action( 'wp_insert_post', [ $this, 'set_post_attributes' ] );
+			wp_update_post( $post, false, false );
+			add_action( 'wp_insert_post', [ $this, 'set_post_attributes' ], 10, 3 );
+			return;
+		}
+
 		if ( ! $update ) {
 			return;
 		}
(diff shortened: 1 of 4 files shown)


includes/publisher/class-publisher-form-editing-experience.php
```diff
@@ -220,7 +220,9 @@ public function enqueue_assets( string $hook ): void {
 						$value = get_post_meta( $post->ID, $field['slug'], true );
 						if ( ! empty( $field['isTitle'] ) && $value !== $post->post_title ) {
 							$post->post_title = $value;
+							remove_action( 'wp_insert_post', [ $this, 'set_post_attributes' ] );
```
nickcernis, line 223: Might it be worth explaining inline why we're removing and re-adding the action here, either as a comment or by extracting this repeated logic as a small function?

nickcernis (commented):
@mindctrl Right now I see the slug change when I change the title field value.

Is that expected? (It's not how post/page title editing works — the slug is constant after publishing unless you explicitly edit it; I can edit the title of a post/page but it's still accessible at the original slug.)

Not a blocker but it might make slugs unstable if queries expect them not to change unexpectedly, so it seemed worth mentioning.

mindctrl:
@nickcernis good catch. Not intended. I think I broke it refactoring some bits before pushing up. Will take another look.

includes/publisher/class-publisher-form-editing-experience.php
mindctrl: Extracted in https://github.com/wpengine/atlas-content-modeler/pull/502/commits/92e4897184116a9727aad9a92bdd3894edc9438b.

matthewguywright (approved):

nickcernis (approved):
Lookin' good! I manually confirmed that:

- [x] Models with a title field copy title to post slug on save.
- [x] Models without a title field use the post ID for the slug.
- [x] Editing an existing post to change the title field content does not change the slug.

PR #511, opened 2022-05-02
MTKA-1313 - Add link for making a new entry in the settings view

anthonyburchell:
## PR adds "New Entry" option to the settings page in the content model dropdown.
Screen Shot 2022-04-29 at 3 47 29 PM

### DIFF
--- a/includes/settings/js/src/components/ContentModelDropdown.jsx
+++ b/includes/settings/js/src/components/ContentModelDropdown.jsx
@@ -61,6 +61,8 @@ export const ContentModelDropdown = ({ model }) => {
 				<a
 					className="new"
 					href={`/wp-admin/post-new.php?post_type=${slug}`}
+					target="_blank"
+					rel="noopener noreferrer"
 					onBlur={() => maybeCloseDropdown(setDropdownOpen, timer)}
 				>
 					{__("Add New Entry", "atlas-content-modeler")}
--- a/includes/settings/js/src/components/ContentModelDropdown.jsx
+++ b/includes/settings/js/src/components/ContentModelDropdown.jsx
@@ -59,7 +59,7 @@ export const ContentModelDropdown = ({ model }) => {
 			</button>
 			<div className={`dropdown-content ${dropdownOpen ? "" : "hidden"}`}>
 				<a
-					className="new"
+					className="add-new-entry"
 					href={`/wp-admin/post-new.php?post_type=${slug}`}
 					target="_blank"
 					rel="noopener noreferrer"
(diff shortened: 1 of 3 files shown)


includes/settings/js/src/components/ContentModelDropdown.jsx
```diff
@@ -58,6 +58,13 @@ export const ContentModelDropdown = ({ model }) => {
 				<Icon type="options" />
 			</button>
 			<div className={`dropdown-content ${dropdownOpen ? "" : "hidden"}`}>
+				<a
+					className="new"
```
mindctrl, line 62: It might be helpful to give this a more descriptive class name, unless we don't use/need it and then we could probably remove it?

includes/settings/js/src/components/ContentModelDropdown.jsx
```diff
@@ -58,6 +58,13 @@ export const ContentModelDropdown = ({ model }) => {
...
+					href={`/wp-admin/post-new.php?post_type=${slug}`}
+					onBlur={() => maybeCloseDropdown(setDropdownOpen, timer)}
+				>
+					{__("New Entry", "atlas-content-modeler")}
```
mindctrl, line 66: All the other items in the dropdown are verbs and/or start with verbs.

```suggestion
					{__("Add New Entry", "atlas-content-modeler")}
```

includes/settings/js/src/components/ContentModelDropdown.jsx
```diff
@@ -58,6 +58,13 @@ export const ContentModelDropdown = ({ model }) => {
 				<Icon type="options" />
 			</button>
 			<div className={`dropdown-content ${dropdownOpen ? "" : "hidden"}`}>
+				<a
+					className="new"
+					href={`/wp-admin/post-new.php?post_type=${slug}`}
```
mindctrl, line 63: Since this takes the user away from the current screen, adding this will make it consistent with how the "Open in GraphiQL" link works, which also takes the user away from the current screen.

```suggestion
					href={`/wp-admin/post-new.php?post_type=${slug}`}
					target="_blank"
					rel="noopener noreferrer"
```

mindctrl (changes requested):

includes/settings/js/src/components/ContentModelDropdown.jsx
anthonyburchell: yeah changed to add-new-entry. Good callout

mindctrl (dismissed):

anthonyburchell:
@mindctrl I forgot the changelog. Noticed we had some missing from the .org changelog and included those in there. If you dont mind the final final review :)

mindctrl:
@anthonyburchell we only update the readme.txt in the release PRs. Due to how that script currently works, it ends up being a hassle to maintain it for every PR.

mindctrl (approved):

PR #532, opened 2022-05-13
MTKA-1436 Repeating Email with Min/Max OR Exact

andrewbotz:
## Description

Add Advanced setting(s) for Email Repeatable that allows for min/max inputs of emails OR exact inputs of emails.

https://wpengine.atlassian.net/browse/MTKA-1436

## Checklist

I have:

- [x] Added Min/Max to email repeatable advanced settings
- [x] Added Exact to email repeatable advanced settings
- [x] Added ability to publish email repeatable with both Min/Max OR Exact email inputs with previously implemented validation (required, domains/subdomains, etc.)

## Testing

1. Create model with Email Repeatable that has advanced settings of Min/Max OR Exact
2. Publish using model with advanced min/max OR exact settings

**Notes**: when publishing, the end-user should only be able to add inputs that fit the criteria. Example, if the max repeatable is 5, then the end-user should not be able to add more inputs than the max. Same scenario for min and exact.

## Screenshots
Screen Shot 2022-05-13 at 9 31 35 AM
Screen Shot 2022-05-13 at 9 31 17 AM
Screen Shot 2022-05-13 at 9 31 46 AM
Screen Shot 2022-05-13 at 9 31 07 AM


includes/settings/js/src/components/fields/EmailSettings.jsx
```diff
@@ -11,7 +11,26 @@ const emailDomainTextArea = css`
...
+	errors,
+	storedData,
+	setValue,
+	getValues,
+	trigger,
+	watch,
+}) => {
+	watch(["minRepeatable", "maxRepeatable", "exactRepeatable"]);
+	const isRepeatable = getValues("isRepeatableEmail");
+	const isMinMaxDisabled = getValues("exactRepeatable") || false;
```
matthewguywright, line 30: You can use these here instead: !!
`const isMinMaxDisabled = !!getValues("exactRepeatable")`

tests/acceptance/AdvancedSettingsCest.php
```diff
@@ -187,4 +187,65 @@ public function i_can_set_allowed_domains_and_subdomains_for_email_fields( \Acce
...
+		$i->fillField( [ 'name' => 'exactRepeatable' ], '2' );
+
+		$i->click( 'button[data-testid="model-advanced-settings-done-button"]' ); // Save Advanced Settings.
+		$i->wait( 1 );
+		$i->click( 'button[data-testid="edit-model-update-create-button"]' ); // Save the field.
+		$i->wait( 1 );
+
+		// Offsets are used here to prevent “other element would receive the click”
+		// due to the “add field” button overlapping the edit button in the center.
+		$i->clickWithLeftButton( '.field-list button.edit', -5, -5 );
```
matthewguywright, line 213: This is interesting.

matthewguywright (approved):

matthewguywright (changes requested):
I see errors being the same across all fields even if just one field has changed. It seems to be putting the same error in all locations.
Screen Shot 2022-05-16 at 11 43 06 AM
.

I am also seeing errors that don't make sense, such as below there is content in the field but it's showing the domain is wrong. The domain is correct but the @ symbol is missing.
Screen Shot 2022-05-16 at 11 46 59 AM

tests/acceptance/AdvancedSettingsCest.php
andrewbotz: Ya I need to shore up some of these e2e tests. This piece was copied from other tests -- it may not be needed anymore.

matthewguywright (approved):

nickcernis (approved):

PR #536, opened 2022-05-19
MTKA-1493: Ensure repeating fields return non-null value

mindctrl:
## Description
Fixes issue where it's possible for a repeating field to break WPGraphQL queries due to returning null when an iterable is expected.

If you edit a model schema and add a new repeating field after entries have been created, and then you run a GraphQL query that contains that field, it's likely the query will be broken until existing posts have been edited.
Error message:
> "User Error: expected iterable, but did not find one for field BusinessDetail.phoneNumbers."

**Steps to reproduce:**
1. Create a new model
2. Add a single entry for that model
3. Edit the model and add a repeating field (configure as not required)
4. Click the Open in GraphiQL link
5. Run the pre-filled query that includes the new repeating field name
6. See error
7. Edit previously-created entry, do not add any values to the repeating field you just added, update the post.
8. Run GraphQL query again
9. See that error is gone

This PR address this issue by ensuring we return an array if the field is repeatable.

Related, our save post logic is saving empty arrays to the database, which fixes the query problem but is not ideal because:
- We can't expect users to edit all of their existing entries to fix queries when they decide to add fields to their schemas.
- We really shouldn't be littering the database with empty values.

This PR address this save post issue by:
- Not saving empty values
- Cleaning up empty values if they exist

https://wpengine.atlassian.net/browse/MTKA-1493

## Checklist

I have:

## Testing
Existing tests continue to pass.

New test passes

### Manual steps to test:
Using a new model, repeat steps 1-5 from above. You should see no errors.

### DIFF
--- a/includes/content-registration/custom-post-types-registration.php
+++ b/includes/content-registration/custom-post-types-registration.php
@@ -14,6 +14,7 @@ use WPGraphQL\Data\Connection\PostObjectConnectionResolver;
 use WPGraphQL\Model\Post;
 use WPGraphQL\Registry\TypeRegistry;
 use WPGraphQL\Data\DataSource;
+use function WPE\AtlasContentModeler\is_field_repeatable;
 
 add_action( 'init', __NAMESPACE__ . '\register_content_types' );
 /**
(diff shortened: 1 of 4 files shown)


includes/content-registration/custom-post-types-registration.php
```diff
@@ -512,30 +513,24 @@ function register_content_fields_with_graphql( TypeRegistry $type_registry ) {
...
+			$field['type']       = $gql_field_type;
 
-			$is_repeatable_text      = ( $field['isRepeatable'] ?? false ) && 'text' === $field['original_type'];
-			$is_repeatable_rich_text = ( $field['isRepeatableRichText'] ?? false ) && $rich_text;
-			$is_repeatable_number    = ( $field['isRepeatableNumber'] ?? false ) && 'Float' === $field['type'];
-			$is_repeatable_date      = ( $field['isRepeatableDate'] ?? false ) && 'date' === $field['original_type'];
-			$is_repeatable_media     = ( $field['isRepeatableMedia'] ?? false ) && 'MediaItem' === $field['type'];
-			$is_repeatable_email     = ( $field['isRepeatableEmail'] ?? false ) && 'email' === $field['original_type'];
-
-			if ( $is_repeatable_text || $is_repeatable_rich_text || $is_repeatable_date || $is_repeatable_email ) {
```
mindctrl, line 525: I refactored all this because:
- I needed to know if the field was repeatable, the type didn't really matter.
- since we mutate the field array (`$field['type'] specifically), using `is_repeatable_field()` can return the wrong boolean, as the default is false and if there's no field type match we potentially get the wrong value.
- passing all these variables to the resolver below is starting to feel gross.

nickcernis (approved):
Working well for me! Thanks for the refactoring too.

```
  "data": {
    "test1493s": {
      "nodes": [
        {
          "regularText": "Test",
          "textRepeater": []
        }
      ]
    }
  },
```

PR #546, opened 2022-05-26
Added validation for repeatable text field

rfmeier:
## Description

This pull request adds validation for repeatable text fields.

## Checklist

I have:

### DIFF
--- a/includes/api/validation-functions.php
+++ b/includes/api/validation-functions.php
@@ -102,7 +102,10 @@ function validate_text_field( $value, array $field ): void {
 		$message = \sprintf( \__( '%1$s must be an array of %2$s', 'atlas-content-modeler' ), $field['name'], $field['type'] );
 
 		validate_array( $value, $message );
-		validate_not_empty( $value, $message );
+
+		if ( is_field_required( $field ) ) {
+			validate_not_empty( $value, $message );
+		}
 	}
 
 	validate_array_of(
--- a/includes/api/validation-functions.php
+++ b/includes/api/validation-functions.php
@@ -99,13 +99,19 @@ function validate_model_field_data( array $model_schema, array $data ) {
 function validate_text_field( $value, array $field ): void {
 	if ( is_field_repeatable( $field ) ) {
 		validate_array( $value, "{$field['name']} must be an array of {$field['type']}" );
-	} else {
-		validate_string( $value, "{$field['name']} must be valid {$field['type']}" );
+		validate_not_empty( $value, "{$field['name']} must be an array of {$field['type']}" );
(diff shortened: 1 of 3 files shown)


includes/api/validation-functions.php
```diff
@@ -98,14 +98,30 @@ function validate_model_field_data( array $model_schema, array $data ) {
...
-	} else {
-		validate_string( $value, "{$field['name']} must be valid {$field['type']}" );
-	}
+		// translators: The name and type of the field.
+		$message = \sprintf( \__( '%1$s must be an array of %2$s', 'atlas-content-modeler' ), $field['name'], $field['type'] );
 
-	if ( is_field_required( $field ) ) {
-		validate_not_empty( $value, "{$field['name']} cannot be empty" );
+		validate_array( $value, $message );
+		validate_not_empty( $value, $message );
```
mindctrl, line 105: Wouldn't this force repeating fields to be required?

includes/api/validation-functions.php
```diff
@@ -98,14 +98,30 @@ function validate_model_field_data( array $model_schema, array $data ) {
...
+			validate_string(
+				$field_value,
+				// translators: The type of the field.
+				\sprintf( \__( 'Field must be valid %s', 'atlas-content-modeler' ), $field['type'] )
+			);
+
+			if ( is_field_required( $field ) ) {
+				validate_not_empty(
+					$field_value,
+					\__( 'Field cannot be empty', 'atlas-content-modeler' )
```
mindctrl, line 120: It might be nice to include the field name here.

includes/api/validation-functions.php
rfmeier: Yes. If you make a require repeatable text field, are not all fields required?

So submitting shouldn't return an error message for the `1` index?
```
[
	'names' => [
		'Peter',
		null,
		'John'
	]
]
```

mindctrl: The diff view is a little weird here for me, but this section of code:
```
	if ( is_field_repeatable( $field ) ) {
		// translators: The name and type of the field.
		$message = \sprintf( \__( '%1$s must be an array of %2$s', 'atlas-content-modeler' ), $field['name'], $field['type'] );

		validate_array( $value, $message );
		validate_not_empty( $value, $message );
	}
```

Just because a field is repeatable doesn't mean it's required. It's still ok to submit no data since repeatable and required are separate options. The `is_field_required()` check below seems to cover it properly. If so, can't the `validate_not_empty()` be removed here and let it catch it below?

mindctrl (approved):

PR #547, opened 2022-05-30
[MTKA 1485] Validate text repeater limits

nickcernis:
## Description

Adds validation for min/max repeatable values in text field entry creation CRUD logic.

https://wpengine.atlassian.net/browse/MTKA-1485

## Checklist

I have:

## Testing

Added tests to verify new `validate_row_count_within_repeatable_limits` function.


includes/api/validation-functions.php
```diff
@@ -490,6 +491,42 @@ function validate_max( $value, int $max, string $message = '' ): void {
...
+ * `maxRepeatable` limits optionally offered by repeater fields.
+ *
+ * @param int   $count The number of rows in repeatable field content.
+ * @param array $field Field data with minRepeatable and maxRepeatable values.
+ *
+ * @throws Validation_Exception Exception when `$count` is outside min/max.
+ *
+ * @return void
+ */
+function validate_row_count_within_repeatable_limits( int $count, array $field ): void {
```
nickcernis, line 505: I passed `$field` here because it simplifies the logic for the caller, but I can pass `$min_repeatable` and `$max_repeatable` if preferred.

nickcernis, line 505: I passed `$field` here because it simplifies the logic for the caller, but I can pass `$min_repeatable` and `$max_repeatable` if preferred.

includes/api/validation-functions.php
```diff
@@ -490,6 +491,42 @@ function validate_max( $value, int $max, string $message = '' ): void {
...
+	if ( \is_numeric( $field['minRepeatable'] ?? '' ) ) {
```
rfmeier, line 506: Looks really good! Only question I have is there a reason why `if ( \is_numeric( $field['minRepeatable'] ?? '' ) ) {` defaults to empty string and `if ( \is_numeric( $field['maxRepeatable'] ?? false ) ) {` defaults to a boolean?

rfmeier (dismissed):

includes/api/validation-functions.php
nickcernis: @rfmeier Thanks for catching that! Fixed in [c7d1cc3](https://github.com/wpengine/atlas-content-modeler/pull/547/commits/c7d1cc32ca078d000b9b84616d4d1b2f5434d1ab).

rfmeier (approved):

includes/api/validation-functions.php
anthonyburchell: Field makes sense to me 👍

anthonyburchell (approved):
Looks great, Nick! 👏

PR #559, opened 2022-06-06
MTKA-1486: CRUD: validate min, max, step for Number field

matthewguywright:
## Description

Updated CRUD number field validation to include min, max, and step validation.

https://wpengine.atlassian.net/browse/MTKA-1486

## Checklist

I have:

- [X] Added an entry to CHANGELOG.md.

## Testing

Use CRUD functions to create a number field with invalid values for advanced settings to see errors.


rfmeier (changes requested):

includes/api/validation-functions.php
```diff
@@ -331,6 +368,29 @@ function validate_email_field( $value, array $field ): void {
...
+ * @param string $message Optional. The error message.
+ *
+ * @throws Validation_Exception Exception when value is invalid.
+ *
+ * @return void
+ */
+function validate_number_type( $value, $number_type, string $message = '' ): void {
+	$message = $message ?: \__( 'Value must be a valid number', 'atlas-content-modeler' );
+
+	if ( $number_type === 'integer' && ! is_int( $value ) ) {
```
nickcernis, line 385: I think this is why the tests are failing. You might have to check for integers in a string-safe way, because WP stores field data as strings:

```suggestion
	if ( $number_type === 'integer' && ! filter_var( $value, FILTER_VALIDATE_INT ) ) {
```

rfmeier: @nickcernis is on the right path here. Since this could be submitted from a form, the value could be a string.

`filter_var` seems to work the best here. Although it does return `1` for `true`.

```
wp> filter_var( "1", FILTER_VALIDATE_INT )
=> int(1)
wp> filter_var( "1.2", FILTER_VALIDATE_INT )
=> bool(false)
wp> filter_var( "a", FILTER_VALIDATE_INT )
=> bool(false)
wp> filter_var( true, FILTER_VALIDATE_INT )
=> int(1)
```

But if `validate_number()` is ran before this, it should catch the possible `true` value.

nickcernis (dismissed):

anthonyburchell (dismissed):
This looks great @matthewguywright ! 🚀

mindctrl:
Just a note for @wpengine/mintaka that @rfmeier is taking over this PR and refactoring based on the other work he's been doing.

nickcernis (approved):

rfmeier (approved):

PR #563, opened 2022-06-07
Change when data is sanitized during crud

rfmeier:
## Description

This pull request moves the `sanitize_fields()` call after data validation. This was/is causing false positives when a field is invalid and not required.

Example: If a number field is not required and has a value of `'not_a_number'`. When the value goes through validation, the value becomes `''`. This is causing the validation against a number to skip. There are other instances of this that will need to be addressed after this fix goes in.

Also added as a utility function `trim_space()`, which is recursively trim white space from string values of a given string or array. This function is ran before validation.


nickcernis (approved):
Makes sense to validate first. 👍

Issue #552, opened 2022-06-01
You cannot register duplicate Types to the Schema


ttstauss:
When creating a model that has two very similar relationship fields set up (i.e. a one to many and a many to many to the same model), graphql returns the following error AFTER the model name is updated:

`You cannot register duplicate Types to the Schema. The Type '[NameOfField]ConnectionEdge' already exists in the Schema. Make sure to give new Types a unique name.`

### To reproduce
1. Spin up a new WordPress.
2. Add the Atlas Content Modeler & WPGraphQL plugins.
3. Set up a Model (Faculty/Faculty Members).
4. Set up another Model (Department/Departments) with a title field).
5. Create two departments.
6. In the Faculty model, add One to Many relationship to departments (called Primary Department as a required field).
7. In the Faculty model, add Many to Many relationship to departments (called Other Departments not required).
8. Create a faculty member and add a Primary department and any number of Other departments.
9. Test the query in GraphiQL. Should work perfectly (department titles are accessible).
10. Go to the Faculty model and update the model Singular Name and Plural Name to Person and People respectively.
11. Go back to GraphiQL and test the query with the new name, will return `null` (if you turn on debugging you see the error from above).

Here's an example query:
```
query getPerson {
  person(id: "8", idType: DATABASE_ID) {
    otherDepartments {
      nodes {
        title
      }
    }
    primaryDepartment {
      node {
        title
      }
    }
  }
}
```
And an example result:
```
{
  "data": {
    "person": {
      "otherDepartments": {
        "nodes": null
      },
      "primaryDepartment": {
        "node": null
      }
    }
  },
  "extensions": {
    "debug": [
      {
        "type": "DUPLICATE_TYPE",
        "message": "You cannot register duplicate Types to the Schema. The Type 'PersonToDepartmentConnectionEdge' already exists in the Schema. Make sure to give new Types a unique name.",
        "type_name": "PersonToDepartmentConnectionEdge",
        "stack": [
          "C:\\Users\\User\\Local Sites\\acm\\app\\public\\wp-content\\plugins\\wp-graphql\\src\\Registry\\TypeRegistry.php:606",
          "C:\\Users\\User\\Local Sites\\acm\\app\\public\\wp-content\\plugins\\wp-graphql\\src\\Registry\\TypeRegistry.php:632",
[...]

PR #566, opened 2022-06-10
MTKA-1542: Ensure relationship fields resolve if model slug mismatches singular label

nickcernis:
## Description

Bug fix for #552 so that relationship fields no longer resolve as null if the model slug does not match the singular label.

https://wpengine.atlassian.net/browse/MTKA-1542

## Checklist

I have:

## Testing

Includes unit test for a GraphQL query with a relationship field on a model with a mismatched singular name and slug.

To test manually, import this blueprint with pre-made relationship fields and connected entries:

```sh
wp acm blueprint import http://contentmodel1.wpengine.com/wp-content/uploads/2022/06/relationships.zip
```

Then make GraphQL queries against the `modela`, `modelb`, `modelc` and `mismatch` post types. You should see data for the relationship fields instead of 'null'.

## Screenshots

| Before | After |
| - | - |
| before | after |


mindctrl (approved):

rfmeier (approved):

anthonyburchell (approved):
Looks good and worked in my manual testing. :D 

Thanks for the blueprint for testing too!

PR #571, opened 2022-06-21
MTKA-1540: acm dashboard refactor

matthewguywright:
## Description
Refactored the code from the hackathon to be up to production ready standards.


mindctrl:
Just a note from standup: we want to merge this into a feature branch before `main`. There's other backend work in flight that will also be merged into the feature branch. When they're together in the feature branch, we can open a PR against `main`.

PR #598, opened 2022-07-18
MTKA-1441: Force unique email

andrewbotz:
## Description

Add the ability to make email type unique via Advanced Settings. This first iteration does not include repeaters and only single email input types.

https://wpengine.atlassian.net/browse/MTKA-1441

## Checklist

- [x] Added ability to set an email type to be unique via the Advanced Settings modal
- [x] Validation on unique email check during `save_post(...)` function
- [x] Debounced validation on inputs after 350MS

## Testing

1. Add new model with email type and enforce unique email validation
2. Create new post with model that includes email with unique validation
3. Attempt to create a new post with the email used/saved from step 2
4. Attempt to update a post with the email used/saved from step 2

## Screenshots

Screen Shot 2022-07-18 at 1 34 11 PM

Screen Shot 2022-07-18 at 1 34 40 PM


includes/publisher/class-publisher-form-editing-experience.php
```diff
@@ -403,6 +404,44 @@ public function save_post( int $post_id, WP_Post $post, bool $update ): void {
...
+				global $wpdb;
+
+				$email_value = $posted_values[ $slug ];
+
+				// Only validate uniqueness of non-repeating email fields for now.
+				if ( ! is_array( $email_value ) ) {
+					// phpcs:disable
+					// A direct database call is the quickest way to query
+					// for unique emails.
+					$identical_emails_query = $wpdb->prepare(
```
nickcernis, line 419: I think we'll need to add a JOIN in the query here to only count emails WHERE post_type = [current model's post type].

Otherwise an email address in another model that has the same field id will falsely trigger a conflict.

The query should look something like this (but we'll need to substitute table names and values in the PHP version):

```
SELECT COUNT(*)
FROM `wp_postmeta`
INNER JOIN wp_posts ON wp_postmeta.post_id = wp_posts.ID
WHERE wp_posts.post_type="modelid"
AND post_id != 1234
AND meta_key = "fieldid"
```

includes/rest-api/routes/validate-field.php
```diff
@@ -0,0 +1,88 @@
...
+	if ( null === $post_id && null === $slug && null === $email ) {
+		return new WP_Error( 'acm_missing_fields', __( 'post_id, slug, and email are required parameters.', 'atlas-content-modeler' ) );
+	}
+
+	global $wpdb;
+
+	// phpcs:disable
+	// A direct database call is the quickest way to query
+	// for unique emails.
+	$identical_email_query = $wpdb->prepare(
```
nickcernis, line 56: Same comment here regarding using JOIN to constrain the check to the current post type.

includes/publisher/class-publisher-form-editing-experience.php
```diff
@@ -403,6 +404,47 @@ public function save_post( int $post_id, WP_Post $post, bool $update ): void {
...
+						"SELECT COUNT(*)
+						FROM `{$wpdb->postmeta}`
+						INNER JOIN wp_posts ON wp_postmeta.post_id = wp_posts.ID
+						WHERE wp_posts.post_type = %s
```
nickcernis, line 423: Looking good, @andrewbotz! I think we just need to use:

- `{$wpdb->postmeta}` in place of `wp_postmeta`
- `{$wpdb->posts}` in place of `wp_posts`

So that it still works if someone isn't using the default wp_posts/wp_postmeta table names.

I can fix this up tomorrow morning if you don't have time to get to it this week, though.

includes/publisher/class-publisher-form-editing-experience.php
```diff
@@ -403,6 +404,47 @@ public function save_post( int $post_id, WP_Post $post, bool $update ): void {
 				}
 			}
 
+			$is_field_unique = get_field_from_slug( $slug, $this->models, $post->post_type )['isUnique'];
```
nickcernis, line 407: This needs a fallback value in case the field does not have the isUnique property:

```suggestion
			$is_field_unique = get_field_from_slug( $slug, $this->models, $post->post_type )['isUnique'] ?? false;
```

Otherwise it produces errors on save for email fields that aren't unique:

Screenshot 2022-08-05 at 09 17 52

nickcernis: I went ahead and fixed this in https://github.com/wpengine/atlas-content-modeler/pull/598/commits/007b2a81ec8fb27cd66ad0fa3bcbf83854dd9bdc. Hope that's ok!

includes/publisher/class-publisher-form-editing-experience.php
nickcernis: Fixed in https://github.com/wpengine/atlas-content-modeler/pull/598/commits/974963a34568105c5e1bae51007741f98b3277f1.

nickcernis (dismissed):
I pushed some tiny changes this morning (hope that's ok!) and I think this is all working well now. Nice work, @andrewbotz! 🥳

includes/rest-api/routes/validate-field.php
```diff
@@ -0,0 +1,92 @@
...
+ */
+function register_rest_routes(): void {
+	register_rest_route(
+		'wpe',
+		'/atlas/validate-unique-email',
+		[
+			'methods'             => WP_REST_Server::EDITABLE,
+			'callback'            => __NAMESPACE__ . '\dispatch_get_validate_unique_email',
+			'permission_callback' => static function () {
+				return current_user_can( 'manage_options' );
```
mindctrl, line 29: @andrewbotz was there a reason to restrict this to this capability? This prevents on-the-fly validation from running for other users, like users with the editor role. I wonder if `edit_posts` is more appropriate?

mindctrl (dismissed):
Looks good and works great. Left a small question, but not a blocker IMO since the `save_post` validation still works.

includes/rest-api/routes/validate-field.php
andrewbotz: Nope, just my lack of expertise. Thanks for catching this, i'll change this and read up on these options a bit more.

nickcernis (approved):

includes/rest-api/routes/validate-field.php
mindctrl: @andrewbotz gotcha, just wanted to be sure I wasn't missing something. There's a handy table of roles/caps here for future reference: https://wordpress.org/support/article/roles-and-capabilities/

PR #599, opened 2022-07-18
Adds description to registered opt in tracking setting

anthonyburchell:
This PR adds a description to the registered settings for opt in tracking. To test, visit your local environment `/wp-json/wp/v2` and search for the `atlas_content_modeler_usage_tracking` setting. :)

### DIFF
--- a/includes/settings/settings-callbacks.php
+++ b/includes/settings/settings-callbacks.php
@@ -188,7 +188,7 @@ function register_plugin_settings(): void {
 		'atlas_content_modeler_usage_tracking',
 		[
 			'type'              => 'string',
-			'description'       => 'Opt into anonymous usage tracking to help us make Atlas Content Modeler better.',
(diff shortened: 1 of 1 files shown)


nickcernis (dismissed):
Lookin' good!

Screenshot 2022-07-19 at 09 46 03

includes/settings/settings-callbacks.php
```diff
@@ -188,6 +188,7 @@ function register_plugin_settings(): void {
 		'atlas_content_modeler_usage_tracking',
 		[
 			'type'              => 'string',
+			'description'       => 'Opt into anonymous usage tracking to help us make Atlas Content Modeler better.',
```
mindctrl, line 191: I guess we should make this translatable.

nickcernis (approved):

mindctrl (approved):

PR #601, opened 2022-07-25
NOJIRA: Confirm paths exist before deleting during blueprint cleanup

nickcernis:
## Description

This fix is designed to address a fatal error affecting all new Atlas blueprint installs.

I haven't been able to reproduce it locally, but attempting to install any blueprint in Portal currently produces the error below during the WP instance setup.

If all looks good we will need to release 0.19.2 today to fix this.

```
#0 /nas/content/live/bpatlasblue840/wp-includes/functions.php(7287): unlink('PK\x03\x04\n\x00\x00\x00\x00\x008V\x92T\x00...')
#1 /nas/content/live/bpatlasblue840/wp-content/plugins/atlas-content-modeler/includes/blueprints/import.php(564): wp_delete_file('PK\x03\x04\n\x00\x00\x00\x00\x008V\x92T\x00...')
#2 /nas/content/live/bpatlasblue840/wp-content/plugins/atlas-content-modeler/includes/wp-cli/class-blueprint.php(235): WPE\AtlasContentModeler\Blueprint\Import\cleanup('PK\x03\x04\n\x00\x00\x00\x00\x008V\x92T\x00...', '/nas/content/li...')
#3 [internal function]: WPE\AtlasContentModeler\WP_CLI\Blueprint->import(Array, Array)
#4 phar:///usr/local/bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Dispatcher/CommandFactory.php(100): call_user_func(Array, Array, Array)
#5 [internal function]: WP_CLI\Dispatcher\CommandFactory::WP_CLI\Dispatcher\{closure}(Array, Array)
#6 phar:///usr/local/bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Dispatcher/Subcommand.php(491): call_user_func(Object(Closure), Array, Array)
#7 phar:///usr/local/bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Runner.php(417): WP_CLI\Dispatcher\Subcommand->invoke(Array, Array, Array)
#8 phar:///usr/local/bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Runner.php(440): WP_CLI\Runner->run_command(Array, Array)
#9 phar:///usr/local/bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Runner.php(1237): WP_CLI\Runner->run_command_and_exit()
[...]

PR #603, opened 2022-07-25
NOJIRA: remove blueprint cleanup step

nickcernis:
Because the step is throwing a fatal error in production, I think it's better to remove it than conditionally run it. That way the plugin behaves the same way everywhere. (The cleanup step just removed the .zip file and manifest; it's ok to leave these in place.)

As @andrewbotz pointed out in https://github.com/wpengine/atlas-content-modeler/pull/601#pullrequestreview-1049563443, we don't have a great way to report back to the user that cleanup failed, and we can't really expect them to remove any blueprint files manually in production.

Previous changelog entry from #601 about preventing fatal errors still stands.

This may also cut down with future maintenance/fatal error issues arising from that step.

### DIFF
--- a/docs/wp-cli/index.md
+++ b/docs/wp-cli/index.md
@@ -19,18 +19,13 @@ Imports an ACM blueprint from a URL or File Path.
 ### Synopsis
 
 ```
-wp acm blueprint import <path> [--skip-cleanup]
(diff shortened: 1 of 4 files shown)


andrewbotz (approved):
Makes sense logically, approving.

PR #614, opened 2022-09-14
Prevent model IDs being created or registered with reserved post type slugs

### LINKED ISSUE #613: Visiting the new post/page screen causes a fatal error if a model exists with an ID of 'type'
### To reproduce
Disable the Classic Editor plugin if active (bug only affects Gutenberg editor pages):

1. Create a model with any singular/plural name and an ID of `type`.
2. Visit any of:
 - `http://example.com/wp-admin/post-new.php`
 - `http://example.com/wp-admin/post-new.php?post_type=page`
 - `http://example.com/wp-json/wp/v2/types/page`

You'll see a fatal error and a notice similar to this one:

```
Notice: Undefined index: id in /Users/username/Sites/local/app/public/wp-content/plugins/atlas-content-
modeler/includes/publisher/class-publisher-form-editing-experience.php on line 148
```

And this notice if your 'type' model has any fields:

```
Notice: Undefined index: id in /Users/username/Sites/local/app/public/wp-content/plugins/atlas-content-
modeler/includes/content-registration/custom-post-types-registration.php on line 67
```

### Expected behavior
- Model IDs of 'type' should not be allowed due to conflicts with existing core REST fields.
- Need a path to handle existing models named 'type'.

### Suspected cause
- ACM includes logic to add `acm_fields` and `acm_related_posts` fields to REST responses via `register_rest_field()` (see below).
- When the ACM model is named 'type', those fields are appended to WP core's 'type' REST field instead of to the 'type' post type.
- When you create a new post/page, the block editor [preloads common data from the REST API](https://github.com/WordPress/WordPress/blob/b089b1d99c167a053b8308c3184b66f9bc9f29ef/wp-includes/block-editor.php#L540).
- That preloaded data [includes a request](https://github.com/WordPress/WordPress/blob/32b92a213aadec521e119c7081526e06e6b97e10/wp-admin/edit-form-blocks.php#L66) to `/wp/v2/types/[$post_type]?context=edit`. 
- That request triggers the ACM callbacks which attempt to retrieve the current post ID.
- New posts don't yet have post IDs, so attempts to access `$post['id']` fail, and 'null' is passed along to functions that require an int, resulting in a fatal error.

https://github.com/wpengine/atlas-content-modeler/blob/4e90f7834008e0a8ffcfbf958e9a0a4e38e96f49/includes/publisher/class-publisher-form-editing-experience.php#L138-L153

https://github.com/wpengine/atlas-content-modeler/blob/4e90f7834008e0a8ffcfbf958e9a0a4e38e96f49/includes/content-registration/custom-post-types-registration.php#L54-L74

### Suggested fix
We probably can't just provide a fallback like `$post['id'] ?? false` because we still want to add `acm_fields` and `acm_related_posts` to that post type's REST fields for other functionality such as related fields.

We may need to prefix any model IDs of 'type' ('acm-type'), prevent 'type' from being used as a model ID, or some combination of those.

### Version information
- Plugin version: 0.21.1
- WordPress version: 6.0.2

Full env info was provided with the original report at https://wordpress.org/support/topic/issue-when-trying-to-create-a-page/.

nickcernis:
Resolves #613:

- Prevents models from being registered if their ID matches post type slugs that WordPress expects to be reserved for its own use (includes 'action', 'author', 'order', 'theme', 'type', 'types').
- Disables any existing models that use those model IDs. This is to prevent unexpected behavior and fatal errors like the [one we saw in WP.org support](https://wordpress.org/support/topic/issue-when-trying-to-create-a-page/).
- Displays a message next to the disabled models:

    Screenshot 2022-09-14 at 20 37 10

- Adds a link to new docs explaining how to migrate the model IDs safely. These docs depend on a WP-CLI command that does not yet exist that I will make a new PR for.

The PR also prevents creation of bad model IDs via the import tool:

Screenshot 2022-09-14 at 14 28 40

And via ACM blueprints:

Screenshot 2022-09-14 at 14 29 05

https://wpengine.atlassian.net/browse/MTKA-1615

## Checklist

I have:

## Testing

Added tests for all new features.

### To test manually

1. Check out the `main` branch and create a model with a singular name and model ID of `type`.
2. You will see a blank page and/or fatal error when attempting to create a new post/page at `/wp-admin/post-new.php?post_type=page`
3. Check out the `fix/prevent-reserved-model-ids` branch.
4. You should be able to create pages again.
5. The Content Models → Models page should show a warning next to the model with the bad ID.

## Documentation Changes

[Preview new docs here](https://github.com/wpengine/atlas-content-modeler/blob/main/docs/help/model-id-conflicts.md).


rfmeier (approved):
Looks good. Manual tests passed.

mindctrl (approved):

PR #619, opened 2022-09-21
Adjust build-readme script to strip existing changelog before appending new one

nickcernis:
## Description

Adjusts `npm run build-readme` to strip the existing changelog in readme.txt if present before appending the new one.

To prevent accidentally doubling up the changelog during the build readme step.

## To test

1. Add some changes to the top of `CHANGELOG.md`.
2. Run `npm run build-readme`.

You should see those changes beneath `== Changelog ==` in `readme.txt`. The changelog should not be repeated


mindctrl (approved):
Works great. Thanks for this!

